PR #3, opened 2017-12-13
Clarified hostname setting options (bug #1070345)

lproven:
I've deleted my old branch with the accidentally reflowed file, & created a fresh one with just the relevant paragraph modified. The diff now is clear.


cwickert:
Looks mostly good.

Please add line breaks after 80 chars.

I wonder if we emphasize that/explain why fqdn is preferred (because it's needed for the SSL certs). But this is not a requirement. Just fix the line breaks and I'll accept this.

cwickert:
I just looked at the current text again and your version is already a great improvement. No need to change anything except from the line breaks.

cwickert:
That's not what I wanted, I wanted to squash and merge

PR #5, opened 2017-12-18
small fixes

MaximilianMeister:
(#SDLKC)

### DIFF
--- a/xml/deployment_cluster_administration.xml
+++ b/xml/deployment_cluster_administration.xml
@@ -48,7 +48,7 @@
      defaults to <filename>$HOME/.kube/config</filename>. To use a different
      location, run
     </para>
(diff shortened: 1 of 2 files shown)


xml/deployment_cluster_administration.xml
```diff
@@ -374,7 +374,7 @@ roleRef:
    installation of PTFs by running:
   </para>
 
-<screen>transactional-update reboot ptf install</screen>
+<screen>transactional-update reboot ptf install <replaceable>rpm ... rpm</replaceable></screen>
```
cwickert: Is this really the file name (as in foo-1.2.3.rpm) or the package name (foo or foo-1.2.3)?

If it is the latter, please change this to &lt;package&gt;&lt;replaceable&gt;package ... package&lt;/replaceable&gt;&lt;/package&gt;

MaximilianMeister: needs to be `.rpm` format, so i guess it can stay as it is

PR #6, opened 2017-12-18
Install openldap on the admin node? [BSC#1072017]

lproven:
Recreated branch because Oxygen reflowed the whole file again.

Simplified the instructions to just say use a different machine -- not specifying OS.

Wrapped instructions at 80char mark.

### DIFF
--- a/xml/deployment_cluster_administration.xml
+++ b/xml/deployment_cluster_administration.xml
@@ -4,61 +4,93 @@
   <!ENTITY % entities SYSTEM "entity-decl.ent">
     %entities;
 ]>
-<chapter version="5.0" xml:id="administration" xmlns="http://docbook.org/ns/docbook"
- xmlns:xi="http://www.w3.org/2001/XInclude" xmlns:xlink="http://www.w3.org/1999/xlink">
- <info><title>Cluster Administration</title>
+<chapter version="5.0" xml:id="administration"
+  xmlns="http://docbook.org/ns/docbook"
+  xmlns:xi="http://www.w3.org/2001/XInclude"
+  xmlns:xlink="http://www.w3.org/1999/xlink">
+ <info>
+  <title>Cluster Administration</title>
   <dm:docmanager xmlns:dm="urn:x-suse:ns:docmanager">
-   <dm:bugtracker/>
+   <dm:bugtracker></dm:bugtracker>
    <dm:translation>yes</dm:translation>
   </dm:docmanager>
  </info>
+
  <sect1 xml:id="auth">
   <title>Authentication and Authorization</title>
-  <para> Role-based access control (RBAC) adds the ability to perform authentication and
-   authorization of activities performed against a &kube; cluster. Authentication is concerned with
-   the <quote>who</quote> and authorization is concerned with the <quote>what</quote>. </para>
-  <para>User information is stored in <phrase role="productname">OpenLDAP</phrase> running in a
-   container on your &productname; &admin_node;. You use standard LDAP administration tools for
-   managing these users. </para>
+  <para>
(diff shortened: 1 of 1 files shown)


xml/deployment_cluster_administration.xml
```diff
@@ -106,26 +106,26 @@ mail: <email address>
...
-     package:
-    </para>
-    <screen>&prompt.root; <command>zypper</command> <option>install openldap2</option></screen>
-    <para>
-     Then generate the password hash:
+     To generate the password hash, on a separate machine, install the 
+     <package>openldap2</package> package:
     </para>
+    <screen>
+     &prompt.root; <command>zypper</command> <option>install openldap2</option>
```
cjschroder: Make screen lines flush left, unless you want leading spaces

xml/deployment_cluster_administration.xml
```diff
@@ -106,26 +106,26 @@ mail: <email address>
...
+     Make sure to replace all the parameters in the template above, indicated as &lt;....&gt;.
+     The mail attribute value is used as the login to &dashboard; and &kube;.
     </para>
     <para>
-     Populate your OpenLDAP server with this LDIF file:
+     Transfer the resulting LDIF file to your CaaSP server, and populate your 
+     OpenLDAP server with this file:
     </para>
     <screen>&prompt.root;<command>ldapadd</command> -H ldap://&lt;ADMIN NODE IP&gt;:389 -ZZ -D cn=admin,dc=infra,dc=caasp,dc=local -w &lt;LDAP ADMIN PASSWORD&gt; -f &lt;LDIF FILE&gt;</screen>
     <para>
```
cjschroder: User-defined variables should use the replaceable tags rather than angle braces.

cjschroder: also, check the style guide for the correct conventions for example IP addresses and host and domain names. Some of them have entities.

xml/deployment_cluster_administration.xml
lproven: Done.

xml/deployment_cluster_administration.xml
lproven: Good catch. This required removing CDATA tags but I have replaced all the &gt; ... &lt; markup with <replaceable>.

Where the instructions referred to the angle brackets, I have replaced it with 

... indicated <replaceable>like this</replaceable>.

I've used entities for admin node, worker node and the product name. Those are all I could identify.

lproven:
Revised according to Carla's suggestions from Wednesday.

cwickert:
I find it impossible to review the changes given that the whole file was rewritten.

lproven:
No it was not intended. If you can tell me how to  reset back to the previous commit, I will redo the corrections Carla called for without re-wrapping.

cwickert:
Any progress here?

xml/deployment_cluster_administration.xml
```diff
@@ -92,55 +94,56 @@
...
-     Make sure to replace all the parameters in the template above, indicated
-     as &lt;....&gt;. The mail attribute value is used as the login to
+     Make sure to replace all the parameters indicated <replaceable>like this</replaceable>
+     in the template above. The mail attribute value is used as the login to
      &dashboard; and &kube;.
     </para>
     <para>
-     Populate your OpenLDAP server with this LDIF file:
+     Transfer the resulting LDIF file to your &productname; server and populate your
```
cwickert: For the record: You don't transfer the file, you use the LDAP client tools to connect to the server.

cwickert:
This was dealt with in e8d44f35cc0117be6ad990b3bcdfcb08904b2b05 in the meantime.

PR #7, opened 2017-12-20
Bug 1070624

lproven:
Cleaned up edit so that there is only a single line altered.


cwickert:
@lproven, please don't forget to the second part of the bug, the admin key extraction.

PR #10, opened 2018-01-22
Bug 1043784 table

lproven:
As requested, I've merged the node-type cells in the 1st column of the table. I've also simplified the layout slightly by replacing empty cells with ``<entry/>'' and fixed the indentation.

Consider this a dummy PR -- do not accept it.

If this looks acceptable, I will move it to the Deployment Guide. Where exactly would you like it to go?

### DIFF
--- a/xml/install_quick.xml
+++ b/xml/install_quick.xml
@@ -12,42 +12,65 @@
  xmlns="http://docbook.org/ns/docbook"
  xmlns:xi="http://www.w3.org/2001/XInclude"
  xmlns:xlink="http://www.w3.org/1999/xlink">
- <?suse-quickstart color="suse"?>
+<?suse-quickstart color="suse"?>
  <title>&instquick;</title>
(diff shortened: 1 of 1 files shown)


cwickert:
Why did you close this? And why did you completely reformat the file again?

lproven:
2 main reasons.

1. As you pointed out, I had inserted the table into the wrong file. The diff was also very messy. I merged the header cells as you requested, checked and validated it, then copied *just* the table into the correct file.

2. Because TomS is teaching me to use git-flow, so I created a new, clean branch with git-flow and inserted the table into the correct file (deployment_installation.xml), meaning a nice clean diff.

Carla reviewed it and picked up on a couple of things. 

That branch is here:

https://github.com/SUSE/doc-caasp/tree/feature/BSC%231043784

cwickert:
I see. But every time you do work in a different branch/PR, we loose the compete review history, which makes it very hard to keep track of things.

PR #20, opened 2018-04-06
Fix broken images in deployment_about.xml

simonflood:


lproven:
Thanks @simonflood! Now done.

PR #21, opened 2018-04-25
Remove remaining mention of ``rebootmgr'' command

lproven:
I missed these lines previously. Now removed. 

It's a small change -- if it looks OK, then please just hit the merge button, and I can close a bug that should not have got this old.


tbazant:
The deletion may not break anything.... :-)

PR #33, opened 2018-06-04
Edit and insert mmeister's proposed changes from https://github.com/S…

lproven:
I have incorporated mmeister's proposed changes, but moved his in-line text hints into separate callouts. If this looks good then I will duplicate the changes into the CaaSP 3 deployment guide too.

### DIFF
--- a/xml/deployment_cluster_administration.xml
+++ b/xml/deployment_cluster_administration.xml
@@ -280,10 +280,10 @@ metadata:
   namespace: <replaceable>applicable namespace</replaceable><co xml:id="co.auth.group.role-namespace"/>
 # This set of rules amounts to "allow all"
 rules:
-- apiGroups: [""]
-  resources: [""]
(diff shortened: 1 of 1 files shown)


cjschroder:
It looks good. I have no changes to suggest.

xml/deployment_cluster_administration.xml
```diff
@@ -309,7 +309,27 @@ roleRef:
      <callout arearefs="co.auth.group.role-namespace">
       <para>
        Namespace the new group should be allowed to access. Use
-        <literal>default</literal> for &kube;' default namespace.
+        <literal>default</literal> for &kube;' default namespace
```
tbazant: Enclose the group of callouts in a &lt;calloutlist> if there is none

lproven: It already is. OK, I will merge it, close the bug and move on. Thank you!

PR #37, opened 2018-06-18
Add missing space to paragraph

r0ckarong:
The automatic replacement for the product string wasn't spaced correctly and looked bad in the flowing text.


cwickert:
Thanks!

PR #40, opened 2018-06-21
Bugfix/bsc#1051838

lproven:
In a comment on Bugzilla, Thorsten Kukuk asked for an example using a password hash rather than plaintext.

This also makes the separate worker-node and admin-node example files more useful.


xml/deployment_installation.xml
```diff
@@ -861,21 +861,19 @@
...
        So, for instance, if your working directory is <filename>~/cloud-config</filename>,
-       then for the admin node, you need:
+       then for the admin node, you need these two files:
       </para>
-      <screen>~/cloud-config/cc-admin/user-data</screen>
-      <para>
-       and
-      </para>
-      <screen>~/cloud-config/cc-admin/meta-data</screen>
+      <screen>
```
cwickert: no line break

xml/deployment_installation.xml
```diff
@@ -861,21 +861,19 @@
...
+~/cloud-config/cc-admin/user-data
+~/cloud-config/cc-admin/meta-data
```
cwickert: no line break

xml/deployment_installation.xml
```diff
@@ -861,21 +861,19 @@
...
+      </screen>
       <para>
        For a &worker_node;, you need:
       </para>
-      <screen>~/cloud-config/cc-worker/user-data</screen>
-      <para>
-       and
-      </para>
-      <screen>~/cloud-config/cc-worker/meta-data</screen>
+      <screen>
```
cwickert: no line break

xml/deployment_installation.xml
```diff
@@ -861,21 +861,19 @@
...
+~/cloud-config/cc-worker/user-data
+~/cloud-config/cc-worker/meta-data
```
cwickert: no line break

xml/deployment_installation.xml
```diff
@@ -899,10 +897,11 @@ network-interfaces: |
 #cloud-config
 debug: True
 disable_root: False
+ssh_deletekeys: False
```
cwickert: Care to elaborate?

xml/deployment_installation.xml
```diff
@@ -899,10 +897,11 @@ network-interfaces: |
...
 ssh_pwauth: True
 chpasswd:
   list: |
-    root:<replaceable>root</replaceable>
+    root:<replaceable>MyPassword</replaceable>
```
cwickert: MY_PASSWORD (replacables are written in caps)

xml/deployment_installation.xml
```diff
@@ -921,10 +920,11 @@ suse_caasp:
...
```
cwickert: MY_PASSWORD

xml/deployment_installation.xml
```diff
@@ -936,20 +936,28 @@ suse_caasp:
...
       </screen>
       <para>
-       To create the ISO image for a worker node, use the same command, but 
-       substitute the name of the folder containing the configuration files for
-       a &worker_node; and title the volume <literal>cc-worker</literal>.
+       To create the ISO image for a worker node on a computer running &opensuse;
+       15, substituting the name of the folder containing the configuration
+       files for a &worker_node; and titling the volume <literal>cc-worker</literal>:
       </para>
+      <screen>
```
cwickert: no line break

xml/deployment_installation.xml
```diff
@@ -936,20 +936,28 @@ suse_caasp:
...
+&prompt.user; sudo mkisofs -output <replaceable>cc-worker.img</replaceable> -volid <replaceable>cidata</replaceable> -joliet -rock <replaceable>cc-worker</replaceable>
```
cwickert: no whitespace after &prompt.user, it's already part of the entity.
tag command as <command>.
wrap the command line with \ if it doesn't fit into a single line in the pdf version.
no line break before the end of the screen

xml/deployment_installation.xml
lproven: Done.

xml/deployment_installation.xml
lproven: Done.

xml/deployment_installation.xml
lproven: Done.

xml/deployment_installation.xml
lproven: Done.

xml/deployment_installation.xml
lproven: I can only find one instance of "MyPassword".

xml/deployment_installation.xml
lproven: No -- I was taking the example files from Bugzilla. 

Googling suggests that "for public instances this should definitely be 'true'". Happy to change it if you think that is preferable.

lproven:
I think I have addressed all your comments now. If there are any outstanding, please tell me.

The build failure seems to be due to some external SLE material but I am not experienced at interpreting Travis output.

xml/deployment_installation.xml
cwickert: The team hasn't agreed on a solution yet.

lproven: OK, fair enough. Are you happy with my amendments to this at present, though, or do you want more changes?

PR #41, opened 2018-06-22
Fix some errors in the English of the SES section (LGHP)

lproven:
Fix the phrasing of the English, and use the existing entity when introducing SES.

### DIFF
--- a/xml/admin_integration.xml
+++ b/xml/admin_integration.xml
@@ -16,9 +16,9 @@
   </dm:docmanager>
  </info>
  <para>
-  &productname; can use &suse;'s another product as storage for
-  containers&mdash;&suse; Enterprise Storage (further called SES). This chapter
-  gives details how to integrate these two products in several ways.
+  &productname; can use another &suse; product as storage for
+  containers&mdash;&ses; (henceforth called SES). This chapter gives details on
+  several ways to integrate these two products.
  </para>
  <sect1 xml:id="integration.prerequisites">
   <title>Prerequisites</title>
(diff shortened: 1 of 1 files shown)


xml/admin_integration.xml
```diff
@@ -30,14 +30,14 @@
...
-     &productname; can access the SES cluster.
+     The &productname; cluster can access the SES cluster.
     </para>
    </listitem>
    <listitem>
     <para>
-     &productname; can communicate with the SES nodes&mdash;master, monitoring
-     nodes, OSD nodes and the meta data server in case you need a shared
-     file system. For more details regarding SES refer to
+     That &productname; can communicate with the SES nodes&mdash;master,
```
cwickert: "That" seems wrong here.

xml/admin_integration.xml
```diff
@@ -30,14 +30,14 @@
...
+     monitoring nodes, OSD nodes and the metadata server, in case you need a
+     shared file system. For more details regarding SES refer to
      <link xlink:href="https://www.suse.com/documentation/ses-4/book_storage_admin/data/book_storage_admin.html">SES
```
cwickert: I think we need an article here "the SES documentation", not just "SES documentation". Please correct me if I'm wrong.

Shouldn't we use an entity for SES anyway? Just saying, that might be better taken care off in a different PR.

xml/admin_integration.xml
lproven: No, it was intentional, but I have revised it to be consistent with the previous line.

xml/admin_integration.xml
lproven: Agreed. Fixed -- I have also moved the link text outside the tag, which is more consistent with the other Guides.

I did bring entities for SES, MicroOS and a few other things across from the release-notes entities file.

PR #44, opened 2018-06-22
Add an example of supplying a pre-hashed password 

lproven:
(Example taken from upstream docs).

As requested in:

https://bugzilla.suse.com/show_bug.cgi?id=1070350


cwickert:
I will not merge this before we have clarification in the bug.

lproven:
I have clarified the relevant paragraph. I will also add a cross-reference to this section from the appendix with an admin-node autoyast file.

PR #46, opened 2018-06-28
Remove mention of installing guest additions

lproven:
As requested in bug comment. 

I can't comply with the full request (i.e. link to § on package installation) because we don't have one, but this removes the negative statement about installing guest additions being difficult.

I think this also makes a case for copying the whole sysreqs § into the QS guide as well. Yes a standalone file would be nicer but it's difficult because of the entities. For now a copy-paste-edit may have to do, I think.

### DIFF
--- a/xml/deployment_sysreqs.xml
+++ b/xml/deployment_sysreqs.xml
@@ -196,13 +196,17 @@
       </listitem>
      </itemizedlist>
      <para>
-      When building a cluster from hypervisor VMs, &suse; recommends building
+      When building a cluster from virtual machines, &suse; recommends building
       nodes from pre-installed disk images, rather than installing new instances
       from an ISO image. 
      </para>
      <para>
-      The pre-installed VM images come with the relevant hypervisor's guest
-      tools pre-installed (where this is applicable).
+      Virtual-disk images are available for Hyper-V, KVM, OpenStack, VMware,
+      and Xen. 
+     </para>
+     <para>
+      These VM images include the relevant hypervisor's guest additions or tools
+      (where this is applicable).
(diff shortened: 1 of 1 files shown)


xml/deployment_sysreqs.xml
```diff
@@ -202,9 +202,7 @@
      </para>
      <para>
       The pre-installed VM images come with the relevant hypervisor's guest
-      tools pre-installed (where this is applicable). It is possible to install
-      these manually, but due to the read-only root filesystem, this is not a
-      trivial task.
+      tools pre-installed (where this is applicable).
```
cwickert: Pre-installed images with pre-installed tools? How about simply "come with the relevant hypervisor's guest tools?"

cwickert:
And what about the second part of the bug?

BTW: Please make sure 
* to mention the bug # in the PR description. Currently one can only derive it from the branch name.
* to mention the bug # in the commit message: "Fix foo (BSC#1234567)"

xml/deployment_sysreqs.xml
lproven: Done.

lproven: I've removed the redundant text. I have also added a single line with a short list of available virtual disk image formats.

The "second part of the bug"? Needing more memory on VMware? This is already addressed in the very next paragraph, on disabling the VMware memory-ballooning feature.

lproven:
I have removed the duplicated text.

If the "second part of the bug" is the over-use of memory on the Admin node, then this is caused by not disabling VMware memory ballooning. This is specifically mentioned already.

PR #47, opened 2018-06-29
Kubernetes Dashboard / Kubectl

r0ckarong:
I created a new page for Kubernetes interaction because I gues there will be more stuff in the future and we needed a place for kubectl and kubernetes-dashboard. 

Then I added more detailed instructions on how to install the heapster requirement for the k8s dashboard and moved that to the "installing software with helm" page.

Updated the title of the guide to the correct "Administration Guide" since it was still Deployment guide.

Added a few screenshots (optimized with the daps optipng run).


lproven:
Looks pretty good. I would suggest wrapping lines at 80 columns though.

cwickert:
This PR should never have been merged.

First of all it was broken due to the misplaced &lt;screen&gt; tag which was fixed in 28160fb.

To make things worse, @lproven added an misspelled entity in 7e78f8f.

Don't get me wrong: There is no problem with messing up a feature branch. That's what PRs and reviews are for. But if it doesn't build, don't merge it. Please validate before commit.

And last but not least: A PR like this one with 35 commits should be squashed. Even if the feature branch was broken at some point in time, that state won't end up in develop. And it makes maintenance easier: Imagine you want to update the chapter on the CaaS 2 branch. You don't want to cherry-pick 35 commits but one.

And if you cannot squash everything into a single commit, think about rebasing (git rebase -i) on the feature branch. Reduce the commits to a meaningful number by topics.

PR #61, opened 2018-07-12
Reword the text about the ``issue-generator'' command & its use

lproven:
As per:
https://bugzilla.suse.com/show_bug.cgi?id=1098775

Removed some info and switches. Add some text from man page. (No man pages are included with MicroOS.) Changed description to match what the command does in MicroOS as opposed to SLE,


xml/admin_administration.xml
```diff
@@ -205,40 +205,45 @@
   <sect2 xml:id="commands.node.managment.issue-generator">
    <title>The <command>issue-generator</command> Command</title>
    <para>
-    The <command>issue-generator</command> creates a volatile temporary
-    <filename>/run/issue</filename> file. The file
-    <filename>/etc/issue</filename> should be a symbolic link to the temporary
-    <filename>/run/issue</filename>.
+    On &suse; &mos;, <filename>/etc/issue</filename> is generated by <command>issue-generator</command>.
```
cwickert: Line break at <= 80 chars please.

xml/admin_administration.xml
```diff
@@ -205,40 +205,45 @@
...
+   </para>
```
cwickert: Do we really need a new para after just once sentence?

xml/admin_administration.xml
```diff
@@ -205,40 +205,45 @@
...
+   <para>The <command>issue-generator</command> program allows &mos; to display
```
cwickert: Use 'command' instead of 'program'.
I don't think we need to mention &mos; again, we just had it in the previous sentence.

xml/admin_administration.xml
```diff
@@ -205,40 +205,45 @@
...
    </para>
-<screen>&prompt.root;<command>issue-generator network <replaceable>remove</replaceable> <replaceable>INTERFACE</replaceable></command></screen>
    <para>
-    The command removes file
-    <filename>/run/issue.d/70-<replaceable>INTERFACE</replaceable>.conf
-    </filename>. The file contains the name of the and escape codes for
-    <command>agetty</command>.
+    Packages should install their configuration files in <filename>/usr/lib/issue.d</filename>.
    </para>
    <para>
```
cwickert: Definitely no new paragraph here, these two sentences belong together. User need to understand the difference between /etc/issue.d and /usr/lib/issue.d. The latter is only for files that are shipped with packages. Changes there would be overwritten by updates, so admins must only make changes or additions in /etc/issue.d.

xml/admin_administration.xml
```diff
@@ -205,40 +205,45 @@
...
+    On &suse; &mos;, <filename>/etc/issue</filename> is generated by <command>issue-generator</command>.
+   </para>
+   <para>The <command>issue-generator</command> program allows &mos; to display
+    a dynamic message above the login prompt, which is generated from the
+    contents of several different files.
    </para>
    <para>
-    You can use the command to prefix all directories and files with a
-    specified prefix (PATH in this example):
+    Files in <filename>/etc/issue.d</filename> override files with the same name in
```
cwickert: I think you should start with mentioning that /etc/issue is only a symbolic link to /run/issue, which is generated dynamically by issue-generator. Thus, users must not touch these two files.

Next you would mention that they should make changes or additions in /etc/issue.d, after that you would continue with /usr/lib/issue.d and eventually mention /run/issue.d.

xml/admin_administration.xml
```diff
@@ -205,40 +205,45 @@
...
+   <para>
+    All configuration files are sorted by their filename in lexicographic order,
+    regardless of which of the directories they reside in.
+   </para>
+   <para>
+    If you run the command without any arguments, all the input files will be
+    applied.
+   </para>
+   <para>
+    Which network interfaces are shown can be configured in the file
```
cwickert: You removed all references to network in the previous text, thus nobody will understand what you are talking about here.

cwickert:
Note that the bug request as about "adding a use-case", not about removing anything.

xml/admin_administration.xml
lproven: Done.

xml/admin_administration.xml
lproven: Done.

xml/admin_administration.xml
lproven: Fair point. Changed.

xml/admin_administration.xml
lproven: Done.

PR #52, opened 2018-07-03
- Setup documentation for CaaSP in the Public Cloud

rjschwei:
Public Cloud setup documentation

PR #62, opened 2018-07-16
Integrate public cloud setup from rjschwei

r0ckarong:
fixed for formatting, closes #52


r0ckarong:
@cwickert As requested formatted with xmlformat and pushed this change.

PR #64, opened 2018-07-23
Add missing Dex port to deployment guide

r0ckarong:


lproven:
Looks like a dup of https://bugzilla.suse.com/show_bug.cgi?id=1098639

I was waiting for info. I will close that bug.

PR #57, opened 2018-07-06
Added Velum settings with LDAP Connectors to Configuration section

danielorf:
Updated documentation to include a section covering the Velum Settings page and a subsection specifically for LDAP Connectors.

PR #67, opened 2018-08-02
Add Velum settings with LDAP Connectors to Configuration section (Rework)

r0ckarong:
Replaces #57 

Ongoing info here: https://trello.com/c/Ipsm0Da9/940-3-external-auth-ldap-documentation


xml/quick_configuration.xml
```diff
@@ -132,14 +127,26 @@
...
        </para>
+      </note>
       </listitem>
      </varlistentry>
      
      <varlistentry>
       <term>Port</term>
        <listitem>
         <para>
+         <!-- What's the default here? -->
```
r0ckarong: @danielorf What is the default combination for the configuration?

xml/quick_configuration.xml
```diff
@@ -156,22 +165,38 @@
...
           </para>
+         </note>
          </listitem>
         </varlistentry>
         
         <varlistentry>
          <term>Anonymous</term>
           <listitem>
            <para>
+            Selects if you wish to perform an anonymous bind with
```
r0ckarong: @danielorf Did I describe this correctly? 
Can you leave DN empty? 
Is it displayed if Anonymous is True?

xml/quick_configuration.xml
```diff
@@ -156,22 +165,38 @@
...
            </para>
           </listitem>
          </varlistentry>
          
          <varlistentry>
+          <!-- Is this field shown when Anonymous is set to True? -->
           <term>DN</term>
            <listitem>
             <para>
+             Distinguished Name used for the authentication.
```
r0ckarong: Can you give me a short description what the required DN fields will look like?
Which ones are required?

xml/quick_configuration.xml
danielorf: Port 636 is default in this implementation.
A little more background:  Port 636 is associated with SimpleTLS and port 389 for StartTLS.  The UI doesn't automatically update the port number based StartTLS status, though it would be a useful feature.

xml/quick_configuration.xml
danielorf: Syntax for a DN (distinguished name) is very flexible.  [Section 4](https://tools.ietf.org/html/rfc4514.html) of this RFC gives some examples.  A common thread among the examples is ",dc=domain,dc=tld" but this can be preceded by many combinations of "cn", "uid" and "ou" values.  It's all up to how the LDAP server is structured and configured by the end user.  Since these LDAP servers will often predate a CaaSP cluster, we must be able to accept the full range allowed by the syntax.

r0ckarong:
Still can't merge this because waiting for this to be added to velum

PR #68, opened 2018-08-09
Rewrite explanation of transactional update snapshots,

r0ckarong:
Add note about manual update race condition
Add TX shell explanation

### DIFF
--- a/xml/admin_administration.xml
+++ b/xml/admin_administration.xml
@@ -54,19 +54,18 @@
    Transactional updates are managed by the <command>transactional-update</command> 
    script, which is called once a day. The script checks if any updates are 
    available. If there are any updates to be applied, a new snapshot of the root 
-   file system is created in the background and is updated using 
-   <command>zypper dup</command>. All updates released to this point are applied.
-   The running filesystem/machine state is left untouched.
+   file system is created in the background and is updated from the release 
+   channels. All updates released to this point are applied. The running 
+   filesystem/machine state is left untouched.
   </para>
   <para>
    The new snapshot, once completely updated, is then marked as active and will 
-   be used as the new default after the next refresh of the system. Only the last
-   created snapshot will be used.
+   be used as the new default after the next reboot of the system.
   </para>
   
   <para>
    &dashboard; will show a list of nodes that have new updates
-   available for use. The cluster administrator then uses &dashboard; to refresh 
+   available for use. The cluster administrator then uses &dashboard; to reboot 
    the nodes to the new snapshots to ensure the health of services and 
    configuration. &dashboard; uses <command>salt</command> to safely disable 
    services on the nodes, apply new snapshots, rewrite configurations and then 
@@ -74,11 +73,12 @@
   </para>
   
   <important>
-   <title>Always Use &dashboard; To Update (Refresh) Nodes</title>
+   <title>Always Use &dashboard; To Update Nodes</title>
   <para>
    It is paramount that you never "hard reboot" nodes in the cluster after 
    transactional updates. This will omit reconfiguring services and applications
-   and will leave nodes in unhealthy, if not unsusable, states. 
+   and will leave nodes in unhealthy, if not unsusable, states. If you have nodes
+   with pending transactional updates follow <xref linkend="transactional.updates.installation"/>.
   </para>
  </important>
    
@@ -95,12 +95,12 @@
   </note>
 
   <sect2 xml:id="transactional.updates.installation">
-   <title>Manually Applying Updates</title>
+   <title>Applying Updates</title>
    
    <para>
     After the <command>transactional-update</command> script has run on all
     nodes, &dashboard; displays any nodes in your cluster running outdated
-    software. The updates are only applied after a refresh. For this purpose,
+    software. The updates are only applied after a reboot. For this purpose,
     &dashboard; enables you to update your cluster directly.
     Follow the next procedure to update your cluster.
    </para>
@@ -115,7 +115,8 @@
      changes will be overwritten by the automatically created snapshot. Moreover, 
      if you perform another manual installation before a refresh, only the last 
      snapshot will be used and all prior changes are lost. Either perform all 
-     updates in one session or refresh the node(s) between each step.
+     updates in one session or reboot the node(s) from &dashboard; between each 
+     step.
   </para>
  </important>
  
--- a/xml/admin_administration.xml
+++ b/xml/admin_administration.xml
@@ -77,7 +77,7 @@
    <title>Always use &dashboard; to update nodes.</title>
   <para>
    It is paramount that you never "hard reboot" nodes in the cluster after 
-   transactional updtes. This will omit reconfiguring services and applications
+   transactional updates. This will omit reconfiguring services and applications
    and will leave nodes in unhealthy, if not unsusable, states. 
   </para>
  </important>
--- a/xml/admin_administration.xml
+++ b/xml/admin_administration.xml
@@ -74,7 +74,7 @@
   </para>
   
   <important>
-   <title>Always use &dashboard; to update nodes.</title>
+   <title>Always Use &dashboard; To Update Nodes</title>
   <para>
    It is paramount that you never "hard reboot" nodes in the cluster after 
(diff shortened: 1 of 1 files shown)


xml/admin_administration.xml
```diff
@@ -75,13 +77,30 @@
...
+   <title>Manually Applying Updates</title>
+   
    <para>
     After the <command>transactional-update</command> script has run on all
     nodes, &dashboard; displays any nodes in your cluster running outdated
-    software. &dashboard; then enables you to update your cluster directly.
+    software. The updates are only applied after a reboot. For this purpose,
+    &dashboard; enables you to update your cluster directly.
     Follow the next procedure to update your cluster.
    </para>
```
ajaeger: What about adding something like: Note that the cluster update via velum might also change configuration settings due to the new software. IF you reboot a machine without doing the orchestration via velum, the configuration update will not happen and this might break your cluster.

Thorsten, @flavio , is this correct?

xml/admin_administration.xml
```diff
@@ -51,17 +51,19 @@
...
-   Transactional updates are managed by the
-   <command>transactional-update</command> script, which is called once a day.
-   The script checks if any updates are available. If there are any updates to
-   be applied, a new snapshot of the root file system is created and the system
-   is updated using the <command>zypper dup</command>. All updates released to
-   this point are applied. The snapshot is then marked as active and will be
-   used after the next reboot of the system. Ensure that the cluster is
-   rebooted as soon as possible after the update installation is complete,
-   otherwise all changes will be lost.
+   Transactional updates are managed by the <command>transactional-update</command>
```
ajaeger: I would add a paragraph here:

Updating is a two step process. First transactional update is run on individual machines to apply the updates for the next boot. Second, the update  needs to be applied. The first step is run automatically every day, the second step is orchestrated from Velum so that any configuration updates can be done as part of the update.

r0ckarong: I've just rewritten a section explaining this.

xml/admin_administration.xml
r0ckarong: This is also covered in the section I've added, rewritten.

xml/admin_administration.xml
```diff
@@ -49,19 +49,39 @@
...
-   is updated using the <command>zypper dup</command>. All updates released to
-   this point are applied. The snapshot is then marked as active and will be
-   used after the next reboot of the system. Ensure that the cluster is
-   rebooted as soon as possible after the update installation is complete,
-   otherwise all changes will be lost.
+   Transactional updates are managed by the <command>transactional-update</command> 
+   script, which is called once a day. The script checks if any updates are 
+   available. If there are any updates to be applied, a new snapshot of the root 
+   file system is created in the background and is updated using 
+   <command>zypper dup</command>. All updates released to this point are applied.
```
thkukuk: I know that it was written in the documentation before, but we should not mention "zypper dup", that's an implementation detail which can always change and should not be from interest for an admin.

xml/admin_administration.xml
```diff
@@ -49,19 +49,39 @@
...
+   The running filesystem/machine state is left untouched.
   </para>
-
+  <para>
+   The new snapshot, once completely updated, is then marked as active and will 
+   be used as the new default after the next refresh of the system. Only the last
```
thkukuk: /refresh/reboot/
I doubt that many people will understand that refresh in this context means reboot.

xml/admin_administration.xml
r0ckarong: Agreed, removing the mention of zypper.

xml/admin_administration.xml
```diff
@@ -75,13 +95,30 @@
...
+     create a snapshot and modify transactional updates, make sure that the 
+     node is rebooted as soon as possible after you are finished. If the 
+     automated transactional update script runs before you reboot, your manual 
+     changes will be overwritten by the automatically created snapshot. Moreover, 
+     if you perform another manual installation before a reboot, only the last 
+     snapshot will be used and all prior changes are lost. Either perform all 
+     updates in one session or reboot the node(s) between each step.
+  </para>
+ </important>
+
```
thkukuk: Above we write, the customer should never do a hard reboot, now we leave him alone and don't tell him, how to trigger the reboot correct.
Else: the title "Manual Transactional Update" does not fit to the description, which describes "transactional-update shell", not how to manual update the system.
Is there really a valid usecase for an admin to know and use this option?

xml/admin_administration.xml
```diff
@@ -49,19 +49,39 @@
...
+   available. If there are any updates to be applied, a new snapshot of the root 
+   file system is created in the background and is updated using 
+   <command>zypper dup</command>. All updates released to this point are applied.
+   The running filesystem/machine state is left untouched.
   </para>
-
+  <para>
+   The new snapshot, once completely updated, is then marked as active and will 
+   be used as the new default after the next refresh of the system. Only the last
+   created snapshot will be used.
```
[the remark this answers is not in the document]
r0ckarong: Removed

xml/admin_administration.xml
r0ckarong: I wouldn't know if there is a use case for manual updates via shell.
@ajaeger Is that supposed to be supported or should we hide this altogether?

@thkukuk The proper way to update the system is described in "Updating the Cluster with Velum." that doesnt' cover what you're asking?

thkukuk:
On Thu, Aug 09, Markus Napp wrote:

> I wouldn't know if there is a use case for manual updates via shell.
> @ajaeger Is that supposed to be supported or should we hide this altogether?
> 
> @thkukuk The proper way to update the system is described in "Updating the
> Cluster with Velum." that doesnt' cover what you're asking?

With "transactional-update shell" you cannot manual update the shell.
In a lot of cases you even don't have network acces, at least not on
CaaSP v1-v3. It's only good to make some changes to the root filesystem,
like creating directories in the else read-only part of the filesystem.
Manual updating the system would be "transactional-update dup reboot"

-- 
Thorsten Kukuk, Distinguished Engineer, Senior Architect SLES & CaaSP
SUSE LINUX GmbH, Maxfeldstr. 5, 90409 Nuernberg, Germany
GF: Felix Imendoerffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nuernberg)

thkukuk:
On Thu, Aug 09, Markus Napp wrote:

>     Manual updating the system would be "transactional-update dup reboot"
> 
> But that also foregoes the graceful update via salt no? So this could break a
> node.

As long as we speak about CaaSP v3 and nothing else: no, this
will not break a node, transactional-update knows better what 
to do to trigger the reboot than an admin knows most of the
time.
That's what we learned with CaaSP v1 and v2.

> What I want to have described is THE proper way to apply updates.

This is the proper way.

-- 
Thorsten Kukuk, Distinguished Engineer, Senior Architect SLES & CaaSP
SUSE LINUX GmbH, Maxfeldstr. 5, 90409 Nuernberg, Germany
GF: Felix Imendoerffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nuernberg)

thkukuk:
On Thu, Aug 09, Andreas Jaeger wrote:

> transactional-update shell is not needed for manual updates, you can run
> "transactional-update update"

And this is how the admin should not do it.

-- 
Thorsten Kukuk, Distinguished Engineer, Senior Architect SLES & CaaSP
SUSE LINUX GmbH, Maxfeldstr. 5, 90409 Nuernberg, Germany
GF: Felix Imendoerffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nuernberg)

xml/admin_administration.xml
flavio: @ajaeger yes, that's correct

r0ckarong:
@thkukuk So the "correct" way I understand would be this:

1. Wait until transactional-update script has updated nodes during daily job and it shows in Velum
2. Reboot via Velum

There are other methods apparently but that is not how we want to communicate this should be done.
We won't describe anything to do with tup shell, manual `dup reboot` or manual "hard reboot".

@ajaeger Please confirm, then I can finalize this.

PR #70, opened 2018-08-09
Add upgrade instructions for CaaSP v3

r0ckarong:


r0ckarong:
This PR is NOT to be merged until there is officially a green light about the published v3 upgrade channel packages. If published before it will break user installations.

r0ckarong:
Preview build can be found here: https://w3.suse.de/~mnapp/book.caasp.admin/cha.deploy.upgrade.html#sec.deploy.upgrade.caasp2

xml/admin_upgrade.xml
```diff
@@ -256,4 +256,213 @@ value: value)}'</command></screen>
...
+     <title>Batch size for upgrade</title>
+     <para>
+      In this example we have limited the number of nodes this step will be 
+      performed on to <literal>10 nodes</literal> at a time.
+     </para>
+      
+      <para>
+      This is a precaution to avoid problems on slower network connections.
+      If you are performing this step on a high bandwidth connection (for 
+      example from within the same datacenter as the cluster), you can raise the
```
kiall: This sentence is close, but not quite right.

    .... If your cluster is hosted within a datacenter with a high bandwidth internet connection, you can increase this number. However, leaving it at the suggested batch size of 10 is recommended .... 

I added a "leave it" recommendation, because, the internet is unpredictable - so its better to limit the amount of nodes updating at once.

xml/admin_upgrade.xml
```diff
@@ -256,4 +256,213 @@ value: value)}'</command></screen>
...
+    </para>
+    <procedure>
+     <step>
+      <para>
+       Log in to &dashboard;.
+      </para>
+     </step>
+     <step>
+      <para>
+       Reboot the Admin node as described in
```
kiall: Can we say "Update the admin node, as described in"

Same below for the remaining nodes

xml/admin_upgrade.xml
r0ckarong: Ok I will adjust that.

xml/admin_upgrade.xml
r0ckarong: Will change this too.

xml/admin_upgrade.xml
```diff
@@ -256,4 +256,213 @@ value: value)}'</command></screen>
...
+    SUSEConnect-0.3.11-3.15.1.x86_64
+jid:
+    20180809103558881056
+retcode:
+    0
+   </screen>
+   
+   <para>
+    If the package version is <literal>0.3.11-3.15.1</literal> (or higher) you
+    have the latest updates from the v2 channel installed.
```
MaximilianMeister: i guess that's a good way to know if you're up-to-date, but only if there's no update to SUSEConnect in the meantime... is there maybe a way to run a suseconnect query that tells if we're on the latest v2 head?

xml/admin_upgrade.xml
```diff
@@ -256,4 +256,213 @@ value: value)}'</command></screen>
...
+    </procedure>
+   </sect2>
+  
+  <sect2 xml:id="sec.deploy.upgrade.caasp2.troubleshooting">
+   <title>Troubleshooting</title>
+   <para>
+    In case the upgrade fails, please perform the support data collection by 
+    running <command>supportconfig</command> on the affected nodes. Provide the 
+    resulting files including the <filename>transactional-update-migration.log</filename>
+    to SUSE Support.
```
MaximilianMeister: would it make sense to advise for a rollback here aside of the supportconfig suggestion?

xml/admin_upgrade.xml
```diff
@@ -259,4 +259,213 @@ value: value)}'</command></screen>
...
+    0
+5f6688bbeac94d2ab5c4330dc7043fb2:
+    Removed symlink /etc/systemd/system/timers.target.wants/transactional-update.timer.
+jid:
+    20180807122220543037
+retcode:
+    0
+    </screen>
+   </sect2>
+
```
ajaeger: Add here something as extra step

"If you use SMT or SUSE Manager for mirroring channels: Mirror the  v3 channels on your server."

xml/admin_upgrade.xml
r0ckarong: @MaximilianMeister Can you provide me with the command needed to do this and some sample output? If no, who can provide this? This was @kiall suggestion to do a simple check if you have the latest packages.

xml/admin_upgrade.xml
r0ckarong: @MaximilianMeister Can you link me to rollback instructions? I thought the same thing when writing this but I don't have any information on how to perform a rollback.

xml/admin_upgrade.xml
MaximilianMeister: maybe `SUSEConnect --status` ? i have to try it on some system

xml/admin_upgrade.xml
```diff
@@ -259,4 +259,218 @@ value: value)}'</command></screen>
    </procedure>
   </sect2>
  </sect1>
+ 
+ <sect1 xml:id="sec.deploy.upgrade.caasp2">
+  <title>Upgrading from &productname; 2</title>
+  <note>
+   <para>
+    If you wish to upgrade to &productname; v3 from v1 you must first perform
```
thkukuk: The title says CaaSP v2, but here we say from v1. As we don't support v1 -> v3, I guess the v1 should be v2.

xml/admin_upgrade.xml
r0ckarong: @kiall I'm ok with this reasoning at this stage but customers will require working rollback for any operation. If we don't know how to do it without breaking stuff, we need to get that figured out. (@ajaeger)

ajaeger: In this case (v2->v3 update), rollback would need to be done for the whole cluster. (CC @thkukuk) and that looks tricky.

In other places we should document rollback.

PR #78, opened 2018-08-24
Add logging chapter

r0ckarong:
Adds logging information


xml/admin_logging.xml
```diff
@@ -0,0 +1,446 @@
...
+
+   <itemizedlist>
+    <listitem>
+     <para>
+      all
+     </para>
+    </listitem>
+    <listitem>
+     <para>
+      garbage
```
ereslibre: Maybe we could include a small comment for each loglevel; I think most of them are self-documenting except `garbage`?

xml/admin_logging.xml
```diff
@@ -0,0 +1,446 @@
...
+
+<screen>
+<command>docker logs -f $(docker ps | grep salt-master | awk '{print $1}')</command>
+   </screen>
+  </sect4>
+
+  <sect4 xml:id="sec.admin.logging.log-levels.salt.debug.minion">
+   <title>Salt Minions</title>
+   <note>
+    <title>Modifications on Salt minions are not persistent</title>
```
ereslibre: I don't think this is very accurate; AFAIK if you create as mentioned a new file `/etc/salt/minion.d/100-debug.conf`, this will be kept even after a new snapshot is created by `transactional-update`.

ereslibre: What I think makes sense to explain is that **any** manual modification to files handled by `salt` (that is: managed by the `salt-minions` in the end) will be lost after salt executes any state that touches those files again; however I don't think this is the case for `/etc/salt/minion.d`.

xml/admin_logging.xml
```diff
@@ -0,0 +1,446 @@
...
+  </sect4>
+  </sect3>
+  </sect2>
+
+  <sect2 xml:id="sec.admin.logging.log-levels.kubernetes">
+   <title>&kube; Log Levels</title>
+   <para>
+    For &kube; our default <literal>loglevel</literal> is <literal>2</literal>
+    <link xlink:href="https://kubernetes.io/docs/reference/kubectl/cheatsheet/#kubectl-output-verbosity-and-debugging">Kubernetes Upstream: Output Verbosity and Debugging</link>.
+    However, this setting can be tweaked by using pillar overrides. If modified,
```
ereslibre: We should explain step by step how this is done, and again probably remove salt slang from here (`pillar override`).

xml/admin_monitoring.xml
```diff
@@ -94,4 +96,69 @@ https://raw.githubusercontent.com/kubernetes/heapster/release-1.3/deploy/kube-co
 <screen>&prompt.root;<command>kubectl apply -f grafana-service.yaml</command></screen>
   </sect2>
  </sect1>
+
+ <sect1 xml:id="sec.admin.monitoring.cluster.health">
+  <title>Cluster Health Checking</title>
+  <para>
+   Although &kube; takes care of a lot of traditional deployment problems
+   automatically, it is considered good practice to monitor the availability
```
ereslibre: Maybe this is a good chance to state the Kubernetes philosophy here: that it's self-healing.

xml/admin_logging.xml
```diff
@@ -0,0 +1,446 @@
...
+  <title>About &productname; Logging</title>
+  <para>
+   Logging across the cluster is done on multiple logical levels. You can think
+   of them as four logical layers (simplified).
+  </para>
+
+   <itemizedlist>
+    <listitem>
+     <para>
+      CaaSP Infrastructure (Salt, Velum, LDAP)
```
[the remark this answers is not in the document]
r0ckarong: the full entity is SUSE CaaS Platform, I thought here that was too long but I'll correct it

xml/admin_logging.xml
```diff
@@ -0,0 +1,446 @@
...
+  </para>
+
+  <sect2 xml:id="sec.admin.logging.salt.orchestration">
+   <title>Salt Orchestration Log</title>
+   <para>
+    The <literal>Salt</literal> orchestration logs contains log entries about
+    orchestration events that have changed the cluster.
+   </para>
+   <para>
+    Orchestration events are
```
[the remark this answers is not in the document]
r0ckarong: Then if I avoid the term "orchestration" as mentioned above how should I call this?

xml/admin_logging.xml
```diff
@@ -0,0 +1,446 @@
...
+ </screen>
+ </sect1>
+
+ <sect1 xml:id="sec.admin.logging.fluentd">
+  <title>External log collection (fluentd)</title>
+  <para>
+   There's a notable exception at infra level, and is that we allow audit logs
+   to be configured from within Velum. This allows the audit logs to be written
+   on the &kube; master nodes at <filename>/var/log/kube-apiserver/audit.log</filename>
+   and an external data collector (continuous reading) like <literal>fluentd</literal>
```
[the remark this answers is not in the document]
r0ckarong: That one came from you :) I just rewrote the word to continuous because that made more sense. Apparently `continuously reading` is the correct thing to write here.

xml/admin_logging.xml
r0ckarong: @ereslibre Where do I find out what these comments should look like? Just in the upstream docs?

xml/admin_logging.xml
r0ckarong: @ereslibre Please provide me with the "step by step"-steps

xml/admin_logging.xml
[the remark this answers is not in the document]
ereslibre: Ah, yeah, that was a `(continue reading)` to continue reading the next paragraph, wasn't intended to be included in the documentation. Can be removed then.

xml/admin_logging.xml
ereslibre: Well, [salt logging level documentation is **not** that useful](https://docs.saltstack.com/en/latest/ref/configuration/logging/#log-levels); `garbage` level is documented as "Even more debugging information"... Maybe we can check salt internally and provide something a little more useful?

xml/admin_logging.xml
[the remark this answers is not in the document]
ereslibre: I don't have a good option to be honest; if we consider "bootstrap", "add node", "remove node", "upgrade cluster" "administrative cluster actions", then maybe we can refer to them as such? If you think exposing the Orchestration slang is fine then I have no objections, just wanted to explicitly state that maybe we could hide that if we had a more clear option.

xml/admin_logging.xml
ereslibre: Will do.

xml/admin_logging.xml
r0ckarong: `Maybe we can check salt internally and provide something a little more useful?` Uhm, and who can/will do this? I don't have time to dig into Salt source code. If this really makes sense I would add this as a separate task for later.

xml/admin_logging.xml
ereslibre: For this to be valid we need to apply https://github.com/kubic-project/velum/pull/640 (fixing bsc#1106193)

xml/admin_logging.xml
ereslibre: I will have a look later.

xml/admin_logging.xml
```diff
@@ -0,0 +1,446 @@
...
+      </para>
+      <itemizedlist>
+       <listitem>
+        <para>
+         Pod (systemd, journald, dmesg)
+       </para>
+       <itemizedlist>
+        <listitem>
+         <para>
+          Application
```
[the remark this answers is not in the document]
r0ckarong: Is there something to say against exposing this information in the architecture overview?

xml/admin_logging.xml
r0ckarong: @ereslibre Can you put how to enable verbosity on a cluster as discussed?

ereslibre: Once https://github.com/kubic-project/velum/pull/640 is merged, they can do the following:

* If it's before the cluster first formation
  * Run this command on the admin node anytime before clicking on "Bootstrap cluster"
    * `docker exec -it $(docker ps | grep dashboard | awk '{print $1}') entrypoint.sh bundle exec rake "velum:create_pillar['kube_log_level', 'desired_loglevel']"`

* If the cluster is already formed, and you want to globally modify the loglevel of all Kubernetes components in all machines
  * Run these commands on the admin node:
    * `docker exec -it $(docker ps | grep dashboard | awk '{print $1}') entrypoint.sh bundle exec rake "velum:create_pillar['kube_log_level', 'desired_loglevel']"`
    * `docker exec -it $(docker ps | grep salt-master | awk '{print $1}') salt-run state.orchestrate orch.kubernetes`

In both cases, `desired_loglevel` refer to [a number in the documented Kubernetes verbosity levels](https://github.com/kubic-project/salt/blob/3dca76e2b679a53e80b4dc99634985bd78cebc40/pillar/params.sls#L60-L69).

Example:

 * `docker exec -it $(docker ps | grep dashboard | awk '{print $1}') entrypoint.sh bundle exec rake "velum:create_pillar['kube_log_level', '7']"`

NOTE: For this verbosity level to actually apply in the cluster, https://github.com/kubic-project/velum/pull/640 needs to get merged.

xml/admin_logging.xml
ereslibre: As mentioned in https://github.com/saltstack/salt/issues/4994#issuecomment-18049810, `garbage` `catches all logging messages, even those not being logged by salt, ie, by external libraries in the current interpreter's scope.`

xml/admin_logging.xml
ereslibre: I can confirm that I created `/etc/salt/minion.d/100-debug.conf` in a 3.0 installation (2016.11.4). Added a repo that upgraded salt to 2018.3.0 to test, and after `transactional-update pkg install salt-2018.3.0-1.4.x86_64` and rebooting the machine, salt got upgraded and I could still see the `/etc/salt/minion.d/100-debug.conf`.

I don't know if this can be encouraged or it could bring some issues in the future that we should clearly explain (@thkukuk can help here).

xml/admin_security.xml
```diff
@@ -15,113 +15,115 @@
...
-   <sect2 xml:id="sec.admin.security.users.import-cert">
-    <title>Importing the LDAP Certificate Locally</title>
-    <para>
-     To be able to perform any LDAP queries, you first need to import the
-     generated LDAP certificate to your local trusted certificate storage. On
-     the &admin_node;, run:
-    </para>
-    <screen>&prompt.root;<command>docker exec -it $(docker ps | grep openldap | \
+  <sect2 xml:id="sec.admin.security.users.import-cert">
+   <title>Importing the LDAP Certificate Locally</title>
```
MaximilianMeister: Minor: could be lowcase `locally`

PR #80, opened 2018-08-31
Deployment on VMware

svenseeberg:
Add sections detailing deployment on VMware ESX and ESXi
* using Virtual Disk Images
* using ISO images for installing

This PR also includes other minor changes around the VMware deployment.


xml/deployment_preparation.xml
```diff
@@ -173,38 +179,65 @@
...
+   <title>Converting Images for &vmware; ESX and ESXi</title>
    <para>
-    Additionally, you need to make a separate copy of the downloaded VM
-    disk image for each VM. We suggest keeping the original download
-    elsewhere and making a fresh copy for each node, naming it
-    appropriately: for example, <literal>caas-admin</literal>,
-    <literal>caas-master</literal>, <literal>caas-worker1</literal>,
-    <literal>caas-worker2</literal> and so on.
+    Downloaded disk images need to be converted for usage with &vmware;
+    ESX and ESXi. Run the following command on the downloaded disk
```
thehejik: Run the following command **on ESXi host** on the downloaded disk... (as the command is available only on ESXi host and not SLE)

xml/deployment_installing_nodes.xml
```diff
@@ -1090,4 +1090,43 @@
...
+   &vmware; tools on all nodes that are included in the package
+   <package>open-vm-tools</package>. Log in on the &admin_node; and
+   execute:
+  </para>
+  <procedure>
+   <step>
+    <para>
+     Install <package>open-vm-tools</package> on all nodes.
+    </para>
+<screen>&prompt.root.admin;<command>docker exec $(docker ps -q --filter name=salt-master) salt -P \
```
thehejik: Maybe we should choose one the method for printing out salt-master container id:
* $(docker ps -q --filter name=salt-master)
* $(docker ps | grep salt-master | awk '{print $1}') ... at line 1122

Both methods are equivalent - I would prefer pure docker way and not docker+grep+awk

xml/deployment_preparation.xml
```diff
@@ -300,37 +338,18 @@ suse_caasp:
     For example, to create the ISO image for an admin node on a computer
     running &opensuse; 42:
    </para>
-   <screen>&prompt.user;sudo genisoimage -output <replaceable>cc-admin.img</replaceable> -volid <replaceable>cidata</replaceable> -joliet -rock <replaceable>cc-admin</replaceable></screen>
+   <screen>&prompt.user;<command>sudo genisoimage -output <replaceable>cc-admin.img</replaceable> -volid <replaceable>cidata</replaceable> -joliet -rock <replaceable>cc-admin</replaceable></command></screen>
```
thehejik: IMO better would be to use .iso extension "cc-admin.iso" + the same for cc-worker image... Line 347
... we have an chapter "Preparing an ISO image" but here you create an image with "img" extension.

xml/deployment_preparation.xml
```diff
@@ -343,7 +362,9 @@ suse_caasp:
     </step>
     <step>
      <para>
-      Attach the <filename>cc-admin.img</filename> disk image as a secondary disk.
+      Attach the <filename>cc-admin.img</filename> image as a
+      secondary disk. For details about preparing the image, see <xref
```
thehejik: Secondary (hard) disk would probably also work but I was always attaching virtual DVD unit with mounted cc-admin|worker.iso file - in the ESX webui there is just a need to activate "Connect.." (on boot) checkbox for DVD unit with the mounted image.

xml/deployment_preparation.xml
```diff
@@ -409,7 +399,9 @@ suse_caasp:
     <step>
      <para>
       Attach the <filename>cc-worker.img</filename> disk image as a
-      secondary disk.
+      secondary disk. For details about preparing the image, see <xref
```
thehejik: same as above (sec. disk vs DVD unit)

xml/deployment_preparation.xml
```diff
@@ -300,37 +338,18 @@ suse_caasp:
...
    </para>
-   <screen>&prompt.user;sudo genisoimage -output <replaceable>cc-admin.img</replaceable> -volid <replaceable>cidata</replaceable> -joliet -rock <replaceable>cc-admin</replaceable></screen>
+   <screen>&prompt.user;<command>sudo genisoimage -output <replaceable>cc-admin.img</replaceable> -volid <replaceable>cidata</replaceable> -joliet -rock <replaceable>cc-admin</replaceable></command></screen>
    <para>
     To create the ISO image for a worker node on a computer running &opensuse;
     15, substituting the name of the folder containing the configuration
     files for a &worker_node; and titling the volume <literal>cc-worker</literal>:
    </para>
-   <screen>&prompt.user;sudo mkisofs -output <replaceable>cc-worker.img</replaceable> -volid <replaceable>cidata</replaceable> -joliet -rock <replaceable>cc-worker</replaceable></screen>
+   <screen>&prompt.user;<command>sudo mkisofs -output <replaceable>cc-worker.img</replaceable> -volid <replaceable>cidata</replaceable> -joliet -rock <replaceable>cc-worker</replaceable></command></screen>
```
thehejik: consider img vs. iso

xml/deployment_installing_nodes.xml
r0ckarong: We use the grep+awk method for most of the other commands. If you do change this to the first version, please let me know so we can check where else this is used.

svenseeberg: @r0ckarong I agree with @thehejik on this point. Do you have an opinion here? I like the suggestion and would replace all grep-awk-based commands with the Salt-only solution.

r0ckarong: I agree that this is the better way to do it. It just returns the necessary info other than producing all the info and cutting it down.
I would want to replace this with: `$(docker ps -q -f name salt-master)` (even shorter)
or equivalents across the guides. Please change it across this PR and I'll adjust it in develop and check my other PRs.

PR #84, opened 2018-09-03
fix cloud-init kernel parameter

fujexo:
According to [cloud-init documentation](
https://git.launchpad.net/cloud-init/tree/doc/sources/kernel-cmdline.txt) the kernel parameter is called `cloud-config-url` or `url` instead of `cloud-init-url`. 

If `cloud-init-url` is used, the file `/etc/cloud/cloud.cfg.d/91_kernel_cmdline_url.cfg` is not created.


r0ckarong:
Thanks!

PR #88, opened 2018-09-13
Add instructions for air gapped deployment/configuration

r0ckarong:
https://w3.nue.suse.com/~mnapp/2019-01-08/book.caasp.deployment/cha.deployment.scenarios.html#sec.deploy.scenarios.airgap

The actual contents are in xml/deployment_scenarios.xml


svenseeberg:
The RMT part looks good to me.

airgap_instructions.adoc
```diff
@@ -0,0 +1,206 @@
+CaaSP is sourced from three types of repositories:
+
+. RPM packages - MicroOS
+.. Software updates on the operating system
+.. Provided by SUSE package repositories
+. Container images - Docker, CRI-O
```
flavio: Do we have to mention docker and cri-o other there? It kinda gives the impression we have two kind of container images based on the engine being used. I would just drop that or add something like "used by the container engine"

airgap_instructions.adoc
```diff
@@ -0,0 +1,206 @@
...
+.. Software updates for the deployments and the CaaSP infrastructure layer
+.. Provided by Docker on dockerhub
+.. Provided by Google on k8s.io
```
flavio: We don't pull anything neither from the docker hub nor from k8s.io. You might mention `registry.suse.com`.

r0ckarong: I'll remove that file that is an outdated notes file

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,888 @@
...
+     </para>
+    </listitem>
+    <listitem>
+     <para>
+      443 HTTPS - RMT Server and Helm Chart Repository Mirror
+     </para>
+    </listitem>
+    <listitem>
+     <para>
+      5000 HTTP - Docker Registry
```
flavio: Question: should we recommend to use a TLS secured docker registry?

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,888 @@
...
+     </para>
+    </listitem>
+   </itemizedlist>
+   <note>
+    <title>Additional Port Configuration</title>
+    <para>
+     If you choose to add more Docker registries to your internal network,
+     these must run on different ports than the standard
+     <literal>5000</literal>.
```
flavio: Another solution is to put something like HAProxy or Apache acting as a reverse proxy that handles vhosts. Should we dig into that? I would keep things easy for the 1st iteration, but it's your call

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,888 @@
...
+   <para>
+    Transferring data from the external network mirror to the internal mirror
+    can be performed in many ways. The most common way is portable storage (USB
+    keys or external hard drives).
+   </para>
+   <para>
+    Sizing of the storage is dependent on the number of data sources that need
+    to be stored. Container images can easily measure several Gigabytes per
+    item; although they are generally smaller for Kubernetes related
+    applications. The overall size of any given RPM repository is at least tens
```
flavio: Wait until you see CAP images :rofl:

xml/deployment_scenarios.xml
r0ckarong: If we know how to do that, of course. The secure option is always preferrable to customers over the unsecured one.

xml/deployment_scenarios.xml
r0ckarong: I assume that people will need more than the suse registry in the long run. So if this makes using it easier, I'm all for it. That's something engineering has to decide with Product management if we want to support that in the long run.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,888 @@
...
+     </para>
+<screen>&prompt.sudo;<command>zypper in helm-mirror</command>
+      </screen>
+      </step>
+      <step>
+       <para>
+      Download all charts from the repository.
+     </para>
+<screen>&prompt.user;<command>helm mirror https://kubernetes-charts.suse.com <replaceable>/charts/</replaceable></command>
+      </screen>
```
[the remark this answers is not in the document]
chentex: @flavio @r0ckarong it's not, we have to provide support for windows too so they can merge that.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,888 @@
...
+     </step>
+     <step>
+      <para>
+       Transfer the contents of <filename>/charts/</filename> to the webserver
+       and unpack it to the desired location. The <filename>index.yaml</filename> file
```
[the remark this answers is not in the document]
r0ckarong: Removed

xml/deployment_scenarios.xml
r0ckarong: We can use the info from Vicentes tests: https://github.com/chentex/registry-testing

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,888 @@
...
+<screen>
+     Container images:
+
+     . Update external container registry mirrors from upstream
+     . Copy registry data onto data storage
+     . Connect storage to internal instance and upload data
+     </screen>
+<screen>
+     Synchronizing contents
+     This procedure illustrates how to synchronize the air-gapped registry with the traditional on-premise one.
```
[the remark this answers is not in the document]
r0ckarong: You tell me, if that's how it will officially work I'll describe that. @chentex what did you tests show for this procedure?

xml/deployment_scenarios.xml
flavio: Yes, I agree people will need more than the SUSE registry. We can definitely provide something to address this need.

Let's keep things simple right now and not dig into that.

xml/deployment_scenarios.xml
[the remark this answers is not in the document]
flavio: > @flavio @r0ckarong it's not, 

Ok, we have to update the instructions to invoke `helm-mirror` instead of what is currently written.

> we have to provide support for windows too so they can merge that.

Do you mean we have to provide some code that handles system wide plugins on windows?

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,888 @@
...
+     <para>
+      The default configuration is not using certificates to secure the
+      communication between this registry and all the nodes communicating with
+      it. There is also no authentication and authorization in place.
+     </para>
+     <para>
+      For production usage, we highly recommend deploying the registry using
+      TLS certificates and implementing authentication and authorization by
+      deploying a Portus instance.
+     </para>
```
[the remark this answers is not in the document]
flavio: I think this can be dropped now.

@chentex proved we do not need to deploy a pull-through cache registry, we can deploy a regular one that uses basic http authentication.

Please rewrite this section with the help of @chentex

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,888 @@
...
+      <para>
+       If you need the ability to store modified container images, you will have
+       to create a new registry that will hold these images. The steps needed
+       to run your own full container image registry are not part of this
+       document.
+      </para>
+       <para>
+       For more information you can refer to:
+       <link xlink:href="https://docs.docker.com/registry/">Docker
+       Registry</link>.
```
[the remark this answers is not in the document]
flavio: This can be dropped, see above.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,888 @@
...
+     <para>
+      There is no need to stop the registry services while doing the backup and
+      restore procedures.
+     </para>
+    </note>
+    <procedure>
+     <title>Update Container Registry Mirror</title>
+     <step>
+      <para>
+       Pull all the desired images
```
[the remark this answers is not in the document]
flavio: Ok, I think we know how to address both points right now.

xml/deployment_scenarios.xml
```diff
@@ -48,16 +48,37 @@
...
+ * Needs secure storage `secure_drive` at least 128GB
+
+ Procedure:
+ * Set up mirror servers on external network
+ * Set up mirror servers on airgapped network
+ * Install RMT on `external1`
+ * Register `external1` against SCC, pull updates
+ * Install RMT on `airgap1`
+ * Install helm-mirror on `external2`
+ * Install helm-mirror on `airgap2`
```
flavio: `helm-mirror` is not needed inside of the airgap network. I would change this line to:

```markdown
* Install skopeo on `external2`
```

xml/deployment_scenarios.xml
```diff
@@ -48,16 +48,37 @@
...
+ * Install webserver on `airgap2`
+ * Install docker-registry on `external2`
+ * Install docker-registry on `airgap2`
+ * Run helm-mirror on `external2` and retrieve all charts
+ * Run helm-mirror on `external2` and retrieve all images
```
flavio: This is not done by `helm-mirror` but via `skopeo sync`.

xml/deployment_scenarios.xml
```diff
@@ -48,16 +48,37 @@
...
+ * Copy all images / charts to `secure_drive`
+ * Connect `secure_drive` to `airgap2` and transfer data
+ * Refresh webserver on `airgap2` if necessary
+
+Using mirrors in CaaSP:
+ * Register CaaSP nodes to RMT on `airgap1`
+ * Configure helm on CaaSP admin node to read from `airgap2`
```
flavio: I'm confused by that. The customer is going to run `helm` to deploy CAP (or other tools) from his own laptop or from some other machine of the airgapped network. Doing that from the admin node is fine, but I don't want to give the impression he mus use this specific node.

I would replace that with something like:

```
* On the machines using the helm client, ensure the helm chart repository running on `airgap2` is added.
```

xml/deployment_scenarios.xml
```diff
@@ -375,6 +396,15 @@
     automatically or can be forced via the CLI. You can then copy this data to
     your trusted storage at any point and update the internal mirror.
    </para>
+   <note>
+    <title>Required Modules</title>
+    <para>
+     &rmtool; provides software in so called modules. You must enable all the
+     default modules for &sle; and in addition the <literal>Containers</literal>
+     module. This module needs to be mirrored inside the airgapped network
+     in order to provide all necessary software for other parts of this scenario.
```
flavio: We don't need the containers module to be mirrored inside of the RMT server of the airgapped network.

Neither `skopeo` nor `helm-mirror` are required inside of the airgappend network

xml/deployment_scenarios.xml
```diff
@@ -478,9 +508,9 @@
      </step>
      <step>
       <para>
-       Transfer the contents of <filename>/charts/</filename> to the webserver
-       and unpack it to the desired location. The <filename>index.yaml</filename> file
-       must be in the root directory for your webserver.
+       Transfer the contents of <filename>/charts/</filename> to the to the
+       desired location on the webserver . The <filename>index.yaml</filename>
+       file must be in the root directory for your webserver.
```
flavio: You can drop the `The <filename>index.yaml</filename> file must be in the root directory for your webserver.` You can have your helm cart repository running under a prefix (eg: `http://web.airgapped.lan/suse-charts/`).

flavio:
I just went through some of the changes that happened since my last review. @chentex, please take a look at the whole document, the results of your research are going  to change quite some parts of this document (we don't need a pull-through registry anymore for example).

chentex:
Thanks @flavio we had a quick sync meeting with @r0ckarong today to explain the changes that will undergo according to my research.

xml/deployment_scenarios.xml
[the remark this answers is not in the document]
chentex: Yes, that is what they told me, that we must provide a way to support all OS that Helm works on, so I need to put some time into it to do it.

xml/deployment_scenarios.xml
r0ckarong: Fixed this in the new version

xml/deployment_scenarios.xml
r0ckarong: This listing is obsolete

xml/deployment_scenarios.xml
r0ckarong: Ok will remove this from the instructions

r0ckarong: the registry package is coming from the containers module.
so we do need it mirrored inside the airgap network otherwise you can't install the registry on the internal mirror.

https://www.suse.com/documentation/sles-15/book_sles_docker/data/preparation.html
https://www.suse.com/documentation/sles-15/book_sles_docker/data/sec_docker_registry_installation.html

I will remove skopeo and helm-mirror requirement from the internal mirror.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1030 @@
...
+      <para>
+       SSH into your mirror server on the <literal>external</literal> network.
+      </para>
+     </step>
+     <step>
+      <para>
+       Download all helm charts from the helm chart repository using
+       <command>helm-mirror</command>.
+      </para>
+<screen>&prompt.user;helm-mirror https://kubernetes-charts.suse.com /chartserver/charts
```
flavio: Wrong url, you have to use the SUSE helm chart. That's our only focus right now

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1030 @@
...
+      </screen>
+     </step>
+    </procedure>
+   </sect3>
+   <sect3>
+    <title>Client Configuration</title>
+    <para>
+     Add the webserver as a repo to <command>helm</command> on the admin node.
```
flavio: Maybe you should say something like "go back to the airgap network", the previous step was done inside of the airgap network.

Plus remember what I wrote above about the usage of helm (the client/server architecture)

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1030 @@
...
+      <para>
+       SSH into the mirror server on the external network.
+      </para>
+     </step>
+     <step>
+      <para>
+      Download all charts from the repository to the filesystem
+      (e.g. <filename>/tmp/charts</filename>).
+      </para>
+<screen>&prompt.user;<command>helm-mirror https://kubernetes-charts.suse.com <replaceable>/tmp/charts</replaceable></command>
```
flavio: Wrong url, you have to use the SUSE one

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1030 @@
...
+     </para>
+     <para>
+      For <literal>--dest-creds</literal> you must use the credentials you
+      created during <xref linkend="sec.deploy.scenarios.airgap.container-registry.mirror" />.
+     </para>
+<screen>&prompt.user;<command>skopeo sync --dest-creds <replaceable>USERNAME:PASSWORD</replaceable> \
+dir:<replaceable>/tmp/skopeodata/registry.suse.com/</replaceable> <replaceable>docker://myregistry.local</replaceable></command>
+      </screen>
+    </step>
+    <step>
```
flavio: I would add another step to remove the `/tmp/skopeodata` directory. This is no longer needed and it's wasting quite some disk space.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1030 @@
...
+
+      </para>
+     </step>
+    </procedure>
+
+    <procedure>
+     <title>Refresh information on the &productname; cluster</title>
+     <step>
+      <para>
+       Update the repository information on the &productname; admin node.
```
flavio: On the node where you are using the helm client.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1030 @@
...
+   </sect3>
+  </sect2>
+
+  <sect2 xml:id="sec.deploy.scenarios.airgap.troubleshooting">
+   <title>Troubleshooting</title>
+   <sect3>
+    <title>Skopeo Fails Because Of Self Signed Certificate</title>
+    <para>
+     If you are using a self-signed certificate for the registry you can use
+     <literal>--dest-tls-verify=false</literal> to skip the integrity check.
```
flavio: Or add it to the system level certificates:

```
sudo cp <cert> /etc/pki/trust/anchors/<cert>
sudo update-ca-certificates
```

flavio: @marcov I think skopeo also has a flag to load the certificates from a specific file (so without having to add them to the system ones)

marcov: That's right, you need to have on the host a copy of the server certificate (i.e. the .crt file).

Then you can use the `--dest-cert-dir` flag instead of `--dest-tls-verify=false`, like this:

```bash
skopeo sync --dest-cert-dir /path/to/the/cert/dir docker://alpine:latest docker://localhost:5000
```

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1030 @@
...
+   </para>
+  </sect3>
+
+   <sect3>
+    <title>Helm Chart and Container Image Mirroring</title>
+    <para>
+     &productname; uses <link xlink:href="https://www.helm.sh/">Helm</link> as
+     one method to install additional software on the cluster. The logic behind
+     this relies on <literal>Charts</literal>, which are configuration files
+     that tell the container engine how to deploy software and its dependencies
```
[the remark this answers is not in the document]
r0ckarong: done

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1030 @@
...
+     </varlistentry>
+     <varlistentry>
+      <term>Internal (Airgapped)</term>
+      <listitem>
+       <para>
+        This machine will host the <literal>&rmtool;</literal> for RPM packages,
+        and <literal>Docker registry</literal> for container images
+        as well as a webserver hosting the <literal>Helm chart repository</literal>
+        files.
+       </para>
```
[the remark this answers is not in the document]
r0ckarong: if it's a separate server the procedure will get slightly more complicated because they have to somehow transfer the data from the secure storage to that server, we're pointing to the SLES docs that explain how to configure a webserver

Is that enough?

xml/deployment_scenarios.xml
r0ckarong: done

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1047 @@
...
+  </sect3>
+
+   <sect3>
+    <title>Helm Chart and Container Image Mirroring</title>
+    <para>
+     &productname; uses <link xlink:href="https://www.helm.sh/">Helm</link> as
+     one method to install additional software on the cluster. The logic behind
+     this relies on <literal>Charts</literal>, which are configuration files
+     that tell &kube; how to deploy software and its dependencies
+     and how to configure it. The actual software installed using this method is
```
hwoarang: I would rephrase that to `how to deploy and configure the software and its dependencies`.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1047 @@
...
+     delivered as <literal>Container images</literal>. The location of the
+     container image is stored inside the Helm chart.
+    </para>
+    <para>
+     Helm charts and container images are provided by SUSE, Docker, and others on
```
hwoarang: heml charts are provided in repositories. Registries normally contain just images. No?

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1047 @@
...
+     Sizing of the storage is dependent on the number of data sources that need
+     to be stored. Container images can easily measure several Gigabytes per
+     item; although they are generally smaller for Kubernetes related
+     applications. The overall size of any given RPM repository is at least tens
+     of Gigabytes. For example: At the time of writing, the package repository
+     for &sls; contains approximately <literal>36GB</literal> of data.
+    </para>
+    <para>
+     The storage must be formatted to a filesystem type supporting files larger
+     than 4GB. We recommend <literal>btrfs</literal>.
```
hwoarang: Not sure if using `btrfs` for the mirror host is sensible. I would suggest `xfs` but overall I think we could possibly avoid recommending a particular FS since it also depends on disk configuration etc

xml/deployment_scenarios.xml
hwoarang: Not sure what's wrong with the url (what is the SUSE helm chart?) but this URL is also referenced below (procedure 2.7)

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1047 @@
...
+      device is mounted at <filename>/mnt/storage</filename>.
+     </para>
+    </step>
+    <step>
+     <para>
+      Transfer the container image data to the trusted storage. This will remove
+      all files and directories that are no longer present on the external host
+      from the trusted storage.
+     </para>
+<screen>&prompt.user;<command>rsync -av /tmp/skopeodata/registry.suse.com/ <replaceable>/mnt/storage/container-images/</replaceable> --delete</command>
```
hwoarang: Maybe worth using `--progress` and dropping `-v` in this and subsequent rsync commands.

xml/deployment_scenarios.xml
r0ckarong: Changed the wording to make this clearer.

xml/deployment_scenarios.xml
r0ckarong: Is xfs natively supported by SLE/opensuse? I don't want to have to add instructions how to get a specific FS working to these instructions. btrfs works out of the box and has >4GB support.

hwoarang: XFS is supported by my point mainly is that the FS chose is relevant to other factors such as RAID configuration, HW specs etc which means that `btrfs` may not be the best choice in all cases. That's why I suggested to either drop the suggestion or add a note that admins should also asses the type of FS they need themselves.

r0ckarong: Ok I'll change the wording. I understand this as a "recommendation" is what we suggest and if you need something else you need to figure that out. It's not "your disk must and can only be formatted to btrfs".

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+   <step>
+    <para>
+     After the synchronization is done, you can remove the <filename>skopeodata</filename>
+     directory.
+    </para>
+<screen>&prompt.user;<command>rm -rf /tmp/skopeodata</command></screen>
+   </step>
+   <step>
+    <para>
+     Replace the charts download URL in the index file for the charts
```
chentex: This whole step is no longer needed, helm-mirror already does everything

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1030 @@
...
+     </step>
+     <step>
+      <para>
+       Create basic authentication credentials for registry. Replace
+       <literal>USERNAME</literal> and <literal>PASSWORD</literal> with
+       proper credentials of your choosing.
+      </para>
+<screen>&prompt.user;<command>mkdir auth</command>
+&prompt.user;<command>docker run --entrypoint htpasswd registry:2.6.2 -Bbn <replaceable>USERNAME PASSWORD</replaceable> > auth/htpasswd</command>
+ </screen>
```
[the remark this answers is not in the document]
hwoarang: I think this is still pending correct? The fact that we assume the user does not have docker running means that the instructions on section `PROCEDURE 2.5: SET UP THE EXTERNAL MIRROR` are not correct because they mention `docker save...`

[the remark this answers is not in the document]
chentex: but that is just the `docker service`. Once teh docker service is running he can run all the docker commands. maybe just moving step 8 to the top, to ensure the client will have docker running.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+       <link xlink:href="https://docs.docker.com/registry/configuration/">Docker
+       Registry: Configuration</link>
+      </para>
+     </step>
+     <step>
+      <para>
+       Start the docker service and enable it at boot time:
+      </para>
+<screen>&prompt.root;<command>systemctl enable --now docker.service</command>
+         </screen>
```
flavio: I agree with @chentex, this step has to be moved upper

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+     <step>
+      <para>
+       Start the registry container.
+      </para>
+<screen>&prompt.user;<command>cd /etc/docker/registry</command>
+&prompt.user;<command>docker run -d -p 5000:5000 --restart=always --name registry \
+-v `pwd`<replaceable>/config.secure.yml</replaceable>:/etc/docker/registry/config.yml \
+-v `pwd`/auth:/auth \
+-v `pwd`/certs:/certs \
+-v <replaceable>/registry</replaceable>:/var/lib/registry sles12/registry:2.6.2</command>
```
flavio: Wrong name of the container image (see above)

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+      <para>
+       SSH into the internal mirror server.
+      </para>
+     </step>
+     <step>
+      <para>
+       Connect the trusted storage to the internal mirror and copy the registry
+       container image to local filesystem.
+      </para>
+<screen>&prompt.user;<command>cp /mnt/storage/registry.tar /etc/docker/registry/registry.tar</command>
```
flavio: There's no need to copy the registry to this location. The user can do a simple docker load pointing straight to the USB disk.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+   </screen>
+     </step>
+     <step>
+      <para>
+       Load the registry container image into the local docker instance.
+      </para>
+<screen>&prompt.user;<command>docker load -i /etc/docker/registry/registry.tar</command>
```
flavio: We have to introduce an additional step that starts and enabled docker at boot time:

```
systemctl enable --now docker
```

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+       <link xlink:href="https://docs.docker.com/registry/configuration/">Docker
+       Registry: Configuration</link>
+      </para>
+     </step>
+     <step>
+      <para>
+       Start the docker service and enable it at boot time:
+      </para>
+<screen>&prompt.root;<command>systemctl enable --now docker.service</command>
+         </screen>
```
flavio: Move it before `docker load` is issued.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+     </step>
+     <step>
+      <para>
+       Start the registry container.
+      </para>
+<screen>&prompt.user;<command>cd /etc/docker/registry</command>
+&prompt.user;<command>docker run -d -p 5000:5000 --restart=always --name registry \
+-v `pwd`<replaceable>/config.secure.yml</replaceable>:/etc/docker/registry/config.yml \
+-v `pwd`/auth:/auth \
+-v `pwd`/certs:/certs \
```
flavio: We are pro users (:smile:), let's mark all the volumes read-only unless we really have to write to them.

```
-v `pwd`<replaceable>/config.secure.yml</replaceable>:/etc/docker/registry/config.yml:ro \
-v `pwd`/auth:/auth:ro \
-v `pwd`/certs:/certs:ro \
```

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+     </step>
+     <step>
+      <para>
+       Start the registry container.
+      </para>
+<screen>&prompt.user;<command>cd /etc/docker/registry</command>
+&prompt.user;<command>docker run -d -p 5000:5000 --restart=always --name registry \
+-v `pwd`<replaceable>/config.airgapped.yml</replaceable>:/etc/docker/registry/config.yml \
+-v `pwd`/auth:/auth \
+-v `pwd`/certs:/certs \
```
flavio: We are pro users (:smile:), let's mark all the volumes read-only unless we really have to write to them.

```
-v `pwd`<replaceable>/config.airgapped.yml</replaceable>:/etc/docker/registry/config.yml:ro \
-v `pwd`/auth:/auth:ro \
-v `pwd`/certs:/certs:ro \
```

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+-v <replaceable>/registry</replaceable>:/var/lib/registry sles12/registry:2.6.2</command>
```
flavio: Wrong container image name, it's going to be something like `registry.suse.com/....`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+       address inside the internal network. This hostname must point to the
+       configured webserver. If you provide the webserver on the internal
+       mirror, we recommend configuring the webserver to be available at
+       <literal>/charts</literal> to avoid confusion with the image registry
+       endpoint. The Helm chart repository would then be accessible at
+       <literal>https://myregistry.local:443/charts/</literal>.
+       </para>
+       <para>
+        You will need the configured hostname to replace the original
+        chart repository address in the index file of your repository.
```
flavio: We have done that earlier when using the `helm-mirror` command. I think it would be nice to tell the customer: look, you have already done that before, see `[here]` (with `here` being a link to the section of the document where we explained that).

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+   </screen>
+   </sect3>
+  </sect2>
+
+  <sect2 xml:id="sec.deploy.scenarios.airgap.update">
+   <title>Updating Registry Mirror And Helm Charts</title>
+   <note>
+    <title>Live Update Of Registry</title>
+    <para>
+     There is no need to stop the Docker registry services while doing the
```
flavio: `s/Docker registry/image container registry/`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+      Connect the trusted storage to the internal mirror.
+     </para>
+    </step>
+    <step>
+     <para>
+      Transfer the container image data to the internal mirror. This will remove
+      all files and directories that are no longer present on the trusted storage
+      from the internal mirror. If you have added any images to the registry
+      manually, please back up these first and restore after the sync from
+      the trusted storage is done.
```
flavio: `If you have added...`: drop that. This internal registry is read-only, there's no way to add stuff to it other than copy low level data straight from the registry storage directory.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+    packages locally for the airgapped environment.
+   </para>
+   <para>
+    &sls; bundles software packages in so called modules. You must enable all
+    the default modules for &sle; and in addition the <literal>Containers</literal>
+    module. All enabled modules need to be mirrored inside the airgapped network
+    in order to provide the necessary software for other parts of this scenario.
+   </para>
+   <para>
+    RMT will provide a repository server that holds the packages
```
flavio: `&rmtool;`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+     All members of the &productname; cluster must be able to communicate with
+     the internal mirror server(s) within the airgapped network.
+    </para>
+    <para>
+     Ports:
+    </para>
+    <itemizedlist>
+     <listitem>
+      <para>
+       80 HTTP - RMT Server and Helm chart repository mirror
```
flavio: `&rmtool;`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+      </para>
+     </listitem>
+     <listitem>
+      <para>
+       443 HTTPS - RMT Server and Helm chart repository mirror
```
flavio: `&rmtool;`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+   <title>Deploying &productname;</title>
+   <para>
+    Use the &productname; <xref linkend="book.caasp.deployment"/> as usual.
+    Some of the considerations below apply; depending of the chosen installation
+    medium.
+    </para>
+   <sect3>
+    <title>Using the ISO</title>
+    <para>
+     From YaST register the node against the RMT server. This will ensure the
```
flavio: `&rmtool;`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+     final node zypper repositories are pointed against RMT, moreover all the
```
flavio: `&rmtool;`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+     available updates are going to be installed (-> no need to run a
+     transactional-update right after the installation -> way faster).
+    </para>
+   </sect3>
+   <sect3>
+    <title>Using AutoYast</title>
+    <para>
+     Ensure the admin node is registered against RMT, that will ensure the
```
flavio: `&rmtool;`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+     nodes that are provisioned by AutoYaST are registered against RMT to have all
```
flavio: `&rmtool;`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+     the updates applied
+    </para>
+   </sect3>
+   <sect3>
+    <title>Using a prebuilt image (eg: KVM, Xen)</title>
+    <para>
+     The node has to be registered against RMT. This should be done in the same
```
flavio: `&rmtool;`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+     way as a regular SLE machine: via SUSEConnect.
```
flavio: Stupid questions:
  * don't we have a macro for `SLE`?
  * `SUSEConnect` (command markup)

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+    </para>
+   </sect3>
+
+   <sect3>
+    <title>Existing unregistered running node</title>
+    <para>
+     Again, they should be using SUSEConnect.
```
flavio: `SUSEConnect`?

xml/deployment_scenarios.xml
flavio: Great, let's update this section of the documentation. I don't want to encourage bad security behaviours

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+     <para>
+      Configure &rmtool; on servers.
+     </para>
+     <para>
+      <xref linkend="sec.deploy.scenarios.airgap.rpm-repository"/>
+     </para>
+    </step>
+    <step>
+     <para>
+      Configure Docker registry on servers.
```
flavio: `s/Docker registry/Image container registry/`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+      </para>
+     </listitem>
+     <listitem>
+      <para>
+       443 HTTPS - RMT Server and Helm chart repository mirror
+      </para>
+     </listitem>
+     <listitem>
+      <para>
+       5000 HTTP - Docker registry
```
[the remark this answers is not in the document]
flavio: `s/Docker registry/Image container registry/`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+      <para>
+       Load the registry container image into the local docker instance.
+      </para>
+<screen>&prompt.user;<command>docker load -i /etc/docker/registry/registry.tar</command>
+       </screen>
+     </step>
+     <step>
+      <para>
+       Create the <filename>/etc/docker/registry/config.airgapped.yml</filename>
+       configuration file for Docker registry.
```
flavio: `s/Docker registry/Image container registry/`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+      Perform the shared update procedure to populate the Helm chart repository
+      and registry services.
+     </para>
+     <para>
+      <xref linkend="sec.deploy.scenarios.airgap.rpm-repository.update"/>
+     </para>
+    </step>
+    <step>
+     <para>
+      Deploy CaaSP platform and configure the nodes to use the respective
```
flavio: Instead of `CaasP` we should use it macro (`&productname`?)

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+    <title>Trusted Storage</title>
+    <para>
+     Transferring data from the external network mirror to the internal mirror
+     can be performed in many ways. The most common way is portable storage (USB
+     keys or external hard drives).
+    </para>
+    <para>
+     Sizing of the storage is dependent on the number of data sources that need
+     to be stored. Container images can easily measure several Gigabytes per
+     item; although they are generally smaller for Kubernetes related
```
flavio: Use the `&kube` macro?

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+     used to download all the images referred to in the Helm charts from their
+     respective registry.
+    </para>
+    <para>
+     Helm charts will be provided to the internal network by a webserver and
+     refer to the container images hosted on the internal registry mirror.
+    </para>
+    <para>
+     Once mirroring is configured, you will not have to modify Dockerfile(s) or
+     Kubernetes manifests to use the mirrors. The requests are passed through
```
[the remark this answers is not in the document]
r0ckarong: Well spotted!

xml/deployment_scenarios.xml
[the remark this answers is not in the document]
r0ckarong: Who is taking care of this? I explain later that they should configure the helm repo server to be available on a different webserver root. Isn't that enough? I could just make this mandatory?
RMT server would use the normal endpoint and helm chart server at /charts or is there a port conflict? I couldn't find any network requirements for RMT or SLE15.

xml/deployment_scenarios.xml
r0ckarong: Yes I still need to work this helm distinction in. The other comment is still open because of that.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1210 @@
...
+      You can modify and store your own container images on the external
+      registry and transfer them with the other container images using the same
+      process. If you need to be able to modify and store container images on
+      the internal network, we recommend to create a new registry that will hold
+      these images. The steps needed to run your own full container image
+      registry are not part of this document.
+     </para>
+      <para>
+      For more information you can refer to:
+      <link xlink:href="https://docs.docker.com/registry/">Docker Registry</link>.
```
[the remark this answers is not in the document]
r0ckarong: I already opened a bug when we started talking about how this must use a separate container.
https://bugzilla.suse.com/show_bug.cgi?id=1117182

xml/deployment_scenarios.xml
r0ckarong: I expect users to do exactly that. We give them a way to load images from some external registry into their internal one so they will use it for anything they need there. I don't know if we should explicitly clarify that this is not supported.

xml/deployment_scenarios.xml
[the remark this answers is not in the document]
flavio: @chentex can you take a look at a deployed RMT machine? I think they are using nginx as web server. It would be just a matter of adding a new vhost or whatever is less intrusive

xml/deployment_scenarios.xml
flavio: `docker load` loads the image from the usb stick into the running docker daemon. There's no need to copy the tarball from the usb to the computer hard drive and then do the `docker load`. You are going to save one step

xml/deployment_scenarios.xml
flavio: I think this is stated somewhere in the docs (that we don't support that). We told them to make the registry read only, we gave them instructions on how to do that. If they do other things it's not our responsibility. This sentence gives the impression we are going to "tolerate" that.

xml/deployment_scenarios.xml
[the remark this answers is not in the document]
chentex: OK, will do that today

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1030 @@
...
+   <note>
+    <title>Mirroring Multiple Image Registries / Chart Repositories</title>
+    <para>
+     You can mirror images and charts from multiple registries in one shared
+     internal registry. Do note, however, that you must manually resolve
+     conflicts due to duplicate charts and container images.
+    </para>
+    <para>
+     We highly recommend running separate helm chart and container registry
+     mirrors for each source registry.
```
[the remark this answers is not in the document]
r0ckarong: Charts from somewhere else would mean also mirroring the referenced images, this is currently not possible on the same machine because the container expects the config in exactly one location that is already taken by our SUSE specific registry

xml/deployment_scenarios.xml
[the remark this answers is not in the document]
chentex: Yes it does provide `htpasswd`

xml/deployment_scenarios.xml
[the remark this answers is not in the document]
flavio: I'm fine with this wording. Marking as resolved.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1177 @@
...
+     Additional mirror registries must be run on separate mirror servers for
+     technical reasons.
+    </para>
+   </note>
+
+   <sect3 xml:id="sec.deploy.scenarios.airgap.container-registry.mirror">
+    <title>Mirror Configuration</title>
+    <para>
+     The container image registry is provided as a container image itself.
+     You must download the registry container from SUSE and run it on the
```
flavio: `s/SUSE/&suse;/` ?

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1177 @@
...
+       Install <filename>helm-mirror</filename> and <filename>skopeo</filename>.
+      </para>
+<screen>&prompt.root;zypper in helm-mirror skopeo
+      </screen>
+     </step>
+     <step>
+      <para>
+       Pull the registry container image from &suse;.
+      </para>
+<screen>&prompt.user;<command>docker pull registry.suse.com/sles12/registry:2.6.2</command>
```
flavio: `<screen>&prompt.user;` should be `<screen>&prompt.root;`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1177 @@
...
+   </screen>
+     </step>
+     <step>
+      <para>
+       Save the pulled image to a <literal>.tar</literal> file.
+      </para>
+<screen>&prompt.user;<command>docker save -o /tmp/registry.tar registry.suse.com/sles12/registry:2.6.2</command>
```
flavio: `<screen>&prompt.user;` should be `<screen>&prompt.root;`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1177 @@
...
+      </screen>
+     </step>
+     <step>
+      <para>
+       Connect the trusted storage to the external mirror. Copy the registry
+       image onto the storage.
+      </para>
+<screen>&prompt.user;<command>cp <replaceable>/tmp/registry.tar</replaceable> <replaceable>/mnt/storage/registry.tar</replaceable></command>
```
flavio: I think we can just write `mv` instead of `cp`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1177 @@
...
+     </step>
+     <step>
+      <para>
+       Create basic authentication credentials for the container image registry.
+      </para>
+      <para>
+       Replace <literal>USERNAME</literal> and <literal>PASSWORD</literal> with
+       proper credentials of your choosing.
+      </para>
+<screen>&prompt.user;<command>mkdir /etc/docker/registry/auth</command>
```
flavio: `<screen>&prompt.user;` should be `<screen>&prompt.root;`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1177 @@
...
+&prompt.user;<command>docker run --entrypoint htpasswd registry.suse.com/sles12/registry:2.6.2 -Bbn <replaceable>USERNAME PASSWORD</replaceable> \
```
flavio: `<screen>&prompt.user;` should be `<screen>&prompt.root;`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1177 @@
...
+       <link xlink:href="https://docs.docker.com/registry/configuration/">Docker
+       Registry: Configuration</link>
+      </para>
+     </step>
+     <step>
+      <para>
+       Start the registry container.
+      </para>
+<screen>&prompt.user;<command>cd /etc/docker/registry</command>
+&prompt.user;<command>docker run -d -p 5000:5000 --restart=always --name registry \
```
flavio: `<screen>&prompt.user;` should be `<screen>&prompt.root;`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1177 @@
...
+       For more details on the configuration, refer to:
+       <link xlink:href="https://docs.docker.com/registry/configuration/">Docker
+       Registry: Configuration</link>
+      </para>
+     </step>
+     <step>
+      <para>
+       Start the registry container.
+      </para>
+<screen>&prompt.user;<command>cd /etc/docker/registry</command>
```
flavio: This line can be now removed, the next command references absolute paths.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1177 @@
...
+      </para>
+<screen>&prompt.root;<command>systemctl enable --now docker.service</command>
+         </screen>
+     </step>
+     <step>
+      <para>
+       Connect the trusted storage to the internal mirror and load the registry
+       container image to the local filesystem.
+      </para>
+<screen>&prompt.user;<command>docker load -i /mnt/storage/registry.tar</command>
```
flavio: `<screen>&prompt.user;` should be `<screen>&prompt.root;`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1177 @@
...
+       For more details on the configuration, refer to:
+       <link xlink:href="https://docs.docker.com/registry/configuration/">Docker
+       Registry: Configuration</link>
+      </para>
+     </step>
+     <step>
+      <para>
+       Start the registry container.
+      </para>
+<screen>&prompt.user;<command>cd /etc/docker/registry</command>
```
flavio: This can be dropped, the next command has absolute paths.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1177 @@
...
+&prompt.user;<command>docker run -d -p 5000:5000 --restart=always --name registry \
```
flavio: `<screen>&prompt.user;` should be `<screen>&prompt.root;`

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1177 @@
...
+   </para>
+   <sect3>
+    <title>Mirror Configuration</title>
+    <procedure>
+     <title>Setting Up Chart Repository Mirror</title>
+     <step>
+      <para>
+       We will re-use the webserver that is shipped with &rmtool; to create
+       a virtual host for the chart repository. The repository will be available
+       at <literal>http://mymirror.local:80/charts</literal>.
```
flavio: This might have to be extended if we prove they are available also over HTTPS (see previous comment).

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1177 @@
...
+     </para>
+    </step>
+    <step>
+     <para>
+     Download all charts from the repository to the filesystem
+     (e.g. <filename>/tmp/charts</filename>).
+     </para>
+     <para>
+      This action will download all charts and overwrite the existing Helm
+      chart repository URL. Replace <literal>https://mymirror.local/charts</literal>
```
flavio: **warning:** right now they doc above states the charts are only served over HTTP. If that's the case then this command has to be changed from `https://` to `http://`.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1177 @@
...
+      with the hostname of the webserver providing the Helm chart repository on
+      the internal network.
+     </para>
+<screen>&prompt.user;<command>mkdir <replaceable>/tmp/charts</replaceable></command></screen>
+<screen>&prompt.user;<command>cd <replaceable>/tmp/charts</replaceable></command></screen>
+<screen>&prompt.user;<command>helm-mirror --new-root-url <replaceable>https://mymirror.local/charts</replaceable> https://kubernetes-charts.suse.com <replaceable>/tmp/charts</replaceable></command></screen>
```
flavio: warning: right now they doc above states the charts are only served over HTTP. If that's the case then this command has to be changed from https:// to http://.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1177 @@
...
+    </para>
+    <itemizedlist>
+     <listitem>
+      <para>
+       80 HTTP - &rmtool; Server and Helm chart repository mirror
+      </para>
+     </listitem>
+     <listitem>
+      <para>
+       443 HTTPS - &rmtool; Server
```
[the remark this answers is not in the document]
chentex: We can, but not out of the box, we have to add extra configuration in the nginx.conf file.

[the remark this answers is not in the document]
flavio: Then let's stick with serving the helm charts over HTTP (because that requires 0 changes to the existing configuration - @chentex  please correct me if I'm wrong).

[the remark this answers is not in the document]
chentex: No correction needed. :+1:

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1171 @@
...
+      Connect the trusted storage to the external mirror.
+     </para>
+    </step>
+    <step>
+     <para>
+      Transfer the container image data to the trusted storage. This will remove
+      all files and directories that are no longer present on the external host
+      from the trusted storage.
+     </para>
+<screen>&prompt.user;<command>rsync -aP /tmp/skopeodata/registry.suse.com/ <replaceable>/mnt/storage/registry/</replaceable> --delete</command>
```
flavio: **DANGER:** this command will fail. We have to rsync the registry storage data to the USB disk, not the `/tmp/skopeodata` (which has just been removed - as it was supposed to happen).

The right command is going to be:

```
<screen>&prompt.user;<command>rsync -aP <replaceable>/var/lib/registry/</replaceable> <replaceable>/mnt/storage/registry/</replaceable> --delete</command>
```

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1171 @@
...
+     </screen>
+    </step>
+    <step>
+     <para>
+      Transfer the helm chart data to the internal mirror. This will remove
+      all charts that do not exist on the trusted storage. If you have added
+      any charts to the location manually, please back up these first and
+      restore after the sync from the trusted storage is done.
+     </para>
+<screen>&prompt.user;<command>rsync -aP <replaceable>/mnt/storage/charts</replaceable> <replaceable>/srv/www/htdocs/charts/</replaceable> --delete</command>
```
flavio: **DANGER:** this command will not work as expected. rsync is a PITA with trailing `/` inside of the source. The command above will lead to the creation of `/srv/www/htdocs/charts/charts/...`.

The right command is:

```
screen>&prompt.user;<command>rsync -aP <replaceable>/mnt/storage/charts/</replaceable> <replaceable>/srv/www/htdocs/charts/</replaceable> --delete</command>
```

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1171 @@
...
+<screen>&prompt.user;<command>skopeo sync --dest-creds <replaceable>USERNAME:PASSWORD</replaceable> \
+dir:<replaceable>/tmp/skopeodata/registry.suse.com/</replaceable> <replaceable>docker://mymirror.local:5000</replaceable></command>
+     </screen>
+   </step>
+   <step>
+    <para>
+     After the synchronization is done, you can remove the <filename>skopeodata</filename>
+     directory.
+    </para>
+<screen>&prompt.user;<command>rm -rf /tmp/skopeodata</command></screen>
```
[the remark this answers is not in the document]
r0ckarong: Yup, good catch.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1171 @@
...
+      </para>
+     </step>
+    </procedure>
+   </sect3>
+
+   <sect3 xml:id="sec.deploy.scenarios.airgap.requirements.network">
+    <title>Networking</title>
+    <para>
+     All members of the &productname; cluster must be able to communicate with
+     the internal mirror server(s) within the airgapped network.
```
hwoarang: It might be obvious but maybe worth pointing out that these ports my be opened on the internal mirror server and on any firewall in between this server and the clients.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1171 @@
...
+     <listitem>
+      <para>
+       5000 HTTPS - Container image registry
+      </para>
+     </listitem>
+    </itemizedlist>
+    <para>
+     The external mirror server must be able to communicate with upstream
+     sources on the internet. The minimal required ports are <literal>80</literal>
+     and <literal>443</literal>.
```
hwoarang: This one in particular reads as if port 80 and 443 needs to be opened on the external mirror whereas it truly means that the outgoing traffic on 80 and 443 must be allowed. Maybe we can clarify that.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1171 @@
...
+           </para>
+          </listitem>
+          <listitem>
+           <para>
+            2 (v)CPU
+           </para>
+          </listitem>
+          <listitem>
+           <para>
+            8 GB RAM
```
hwoarang: nit: isn't 8G a bit too much for a node which only runs a registry and RMT? :)

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1171 @@
...
+     of Gigabytes. For example: At the time of writing, the package repository
+     for &sls; contains approximately <literal>36 GB</literal> of data.
+    </para>
+    <para>
+     The storage must be formatted to a filesystem type supporting files larger
+     than <literal>4 GB</literal>.
+    </para>
+    <para>
+     The type of filesystem used for the storage depends on your requirements.
+     We recommend <literal>btrfs</literal>.
```
hwoarang: honestly the `btrfs` suggestion still bugs me especially when talking about external/usb storage.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1171 @@
...
+           </para>
+          </listitem>
+          <listitem>
+           <para>
+            8 GB RAM
+           </para>
+          </listitem>
+          <listitem>
+           <para>
+            500 GB Storage
```
hwoarang: Here we suggest 500G of storage but later on why recommend at least 128G for the external storage. My understanding is that the external storage is suppose to transfer all the useful data from the external mirror to the internal one so if >=128G is all we need, then perhaps we can tune down the disk reqs on the mirror as well.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1171 @@
...
+   <title>RPM Repository Mirror</title>
+   <sect3 xml:id="sec.deploy.scenarios.airgap.rpm-repository.mirror">
+    <title>Mirror Configuration</title>
+    <note>
+     <title>Deploy The Mirror Before &productname; Cluster Deployment</title>
+     <para>
+      The mirror on the airgapped network must be running and populated before
+      deploying &productname;. Starting with &productname; v4 many components
+      will no longer be shipped with the install media and must be fetched
+      from the repository. For v3 this significantly simplifies deployment since
```
hwoarang: Should this say 'v4' ?

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1171 @@
...
+    </note>
+
+    <procedure>
+     <title>Set Up The External Mirror</title>
+     <step>
+      <para>
+       SSH into the external mirror server.
+      </para>
+     </step>
+     <step>
```
hwoarang: should we mention that a `zypper -n in docker` install is needed in order to ensure that docker is installed before we try to start it?? Or better yet move step#3 here and add the `docker` package there.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1171 @@
...
+    </para>
+
+    <procedure>
+     <title>Set Up Internal Mirror</title>
+     <step>
+      <para>
+       SSH into the internal mirror server.
+      </para>
+     </step>
+     <step>
```
hwoarang: same here.

xml/deployment_scenarios.xml
r0ckarong: I have no idea. Who can tell me?

xml/deployment_scenarios.xml
r0ckarong: Then what is the recommended filesystem? As said before I don't want to have to explain how to install ntfs drivers or some other stuff. We need >4GB binary support. We can just recommend ext4? @flavio ?

xml/deployment_scenarios.xml
r0ckarong: Again, I don't really know. If each OS repository has like 30-40GB of data and you download a bunch of container images, you will run out of space quickly. For example you run this operation and keep the results for the mirror in a specific folder for rollback. You have this data 2-3-5-7 etc. times. So you quickly run into hundreds of Gigs of storage needed but the amount you transfer to the internal mirror is much smaller.

That's my perspective though, we haven't done any benchmarking of this use case to tell us how much data actually gets downloaded/created from this.

xml/deployment_scenarios.xml
r0ckarong: Is v4 released? When will it be released? Is this feature only contained in v4? So far I wasn't told that this is separate. If we need to release this as v4 only there is lots of other work that needs to go into docs before we can do this.

hwoarang: This paragraph already mentions `v4` a few lines above and I am not sure why it's there. But the simplification that's mentioned in the last sentence probably affects >= `v4` instead of `v3` that's why I mentioned that here. @flavio ?

xml/deployment_scenarios.xml
flavio: It's probably a bit too much, but honestly I wouldn't bother. If you really want to go the extra mile you look at what are the requirements on the official RMT documentation

xml/deployment_scenarios.xml
flavio: Yes, this is going to affect releases of CaaSP after v3. So v4 will be the first one; it hasn't been released yet.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1175 @@
...
+      <para>
+       Make sure you have
+       <link xlink:href="https://www.suse.com/documentation/sles-15/book_sles_docker/data/preparation.html">enabled
+       the <filename>Containers</filename> module</link> on each server.
+      </para>
+     </step>
+     <step>
+      <para>
+       You will need SSL/TLS certificates to secure services on each server.
```
saschagrunert: From a customer perspective I would not be sure what I have to do here. Are these certs self-signed or do I have to possibility to obtain them from some already deployed resource?

xml/deployment_scenarios.xml
hwoarang: I would say that you don't really have to recommend any fs. If you recommend one, then people might start asking on how to configure it properly. I would leave this as an exercise to the deployer.

xml/deployment_scenarios.xml
hwoarang: OK no problem. My question was mostly because of the difference between 128G and 500G in the recommendations.

xml/deployment_scenarios.xml
hwoarang: OK fine by me.

xml/deployment_scenarios.xml
hwoarang: So we need to replace `For v3 this...` with `For v4 this...` or drop the entire comment about the upcoming v4 release.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1175 @@
...
+       Start the docker service and enable it at boot time:
+      </para>
+<screen>&prompt.root;<command>systemctl enable --now docker.service</command>
+         </screen>
+     </step>
+     <step>
+      <para>
+       Pull the registry container image from &suse;.
+      </para>
+<screen>&prompt.root;<command>docker pull registry.suse.com/sles12/registry:2.6.2</command>
```
saschagrunert: I cannot pull this image, will it be published later on?

```
sascha@sgrunert-test:~> docker pull registry.suse.com/sles12/registry:2.6.2
Error response from daemon: manifest for registry.suse.com/sles12/registry:2.6.2 not found
```

xml/deployment_scenarios.xml
r0ckarong: Removed the comment

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1175 @@
...
+     all the available updates are going to be installed (-> no need to run a
+     transactional-update right after the installation -> way faster).
+    </para>
+   </sect3>
+   <sect3>
+    <title>Using AutoYast</title>
+    <para>
+     Ensure the admin node is registered against &rmtool;, that will ensure the
+     nodes that are provisioned by AutoYaST are registered against &rmtool; to
+     have all the updates applied
```
saschagrunert: There is a missing dot at the end of the line. :innocent:

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1175 @@
...
+    Some of the considerations below apply; depending of the chosen installation
+    medium.
+    </para>
+   <sect3>
+    <title>Using the ISO</title>
+    <para>
+     From YaST register the node against the &rmtool; server. This will ensure
+     the final node zypper repositories are pointed against &rmtool;, moreover
+     all the available updates are going to be installed (-> no need to run a
+     transactional-update right after the installation -> way faster).
```
saschagrunert: I'd suggest to write out what the arrows mean or use unicode arrows → :)

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1175 @@
...
+     <imagedata fileref="direct_connection.png" width="100%"/>
+    </imageobject>
+    <imageobject role="html">
+     <imagedata fileref="direct_connection.png" width="100%"/>
+    </imageobject>
+   </mediaobject>
+  </informalfigure>
+ </sect1>
+ <sect1 xml:id="sec.deploy.scenarios.airgap">
+  <title>Airgapped Deployment</title>
```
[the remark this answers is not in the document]
r0ckarong: Corrected to Air Gapped/air gapped

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1175 @@
...
+  <important>
+   <title>Scope Of This Document</title>
+   <para>
+    This document focuses on providing mirrors for the resources provided by
+    &suse; and required for basic &productname; functionality. If you require
```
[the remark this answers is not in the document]
r0ckarong: the sentence reads 

"and required <CaaSP> functionality".

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1175 @@
...
+    Once the replacement sources are in place, the key components are
+    reconfigured to use the mirrors as their main sources.
+   </para>
+  </sect3>
+
+  <sect3>
+   <title>RPM Package Repository Mirroring</title>
+   <para>
+    Mirroring of the RPM repositories is handled by the
+    <link xlink:href="https://www.suse.com/documentation/sles-15/book_rmt/data/book_rmt.html">&rmtool;</link>
```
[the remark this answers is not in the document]
r0ckarong: Ah, very good call. The entity that replaces this is still wrong. Fixed now.

xml/deployment_scenarios.xml
```diff
@@ -0,0 +1,1175 @@
...
+
+  <sect2 xml:id="sec.deploy.scenarios.airgap.requirements">
+   <title>Requirements</title>
+   <sect3 xml:id="sec.deploy.scenarios.airgap.requirements.machines">
+    <title>Mirror Servers</title>
+    <note>
+     <title>Shared Mirror Server</title>
+     <para>
+      If you have multiple &productname; clusters or a very large amount of
+      nodes accessing the mirrors, you should increase the sizing of CPU/RAM.
```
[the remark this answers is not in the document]
r0ckarong: this is talking about the mirror server. If you have hundred of nodes pulling from the container image repository you might need a beefier machine but I have no idea where I would get numbers for that.

xml/deployment_scenarios.xml
r0ckarong: The assumption is air gapped networking is done in an enterprise network. It's the network admins' job to create and provide certificates.

I don't know if there is somebody out there who cares enough about security to want an air gapped network and then use self-signed certs.

xml/deployment_scenarios.xml
jordimassaguerpla: It is now

xml/deployment_scenarios.xml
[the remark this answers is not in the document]
flavio: I would leave that generic right now.

xml/deployment_scenarios.xml
flavio: I agree with @hwoarang

r0ckarong: I removed the suggestion for FS type.

PR #97, opened 2018-09-28
Add salt cluster sizing requirements and adjustment (bsc#1089303)

r0ckarong:
https://w3.nue.suse.com/%7Emnapp/2018-09-28/book.caasp.admin/cha.admin.html#sec.admin.salt.worker_threads

https://w3.nue.suse.com/~mnapp/2018-09-28/book.caasp.deployment/cha.deploy.requirements.html#sec.deploy.requirements.system.cluster.salt_cluster_size

### DIFF
--- a/xml/admin_administration.xml
+++ b/xml/admin_administration.xml
@@ -138,12 +138,13 @@ cmd.run "<replaceable>df -h</replaceable>"</command>
    The output will be produced in your current terminal session.
   </para>
 
-  <sect2 xml:id="sec.admin.salt.workers">
-   <title>Adjusting The Number Of Salt Workers</title>
+  <sect2 xml:id="sec.admin.salt.worker_threads">
+   <title>Adjusting The Number Of Salt Worker Threads</title>
    <para>
     It will sometimes be necessary to resize the cluster to adjust for workloads
     or other factors. Salt will run into problems, if the number of nodes to
-    handle becomes too large without adjusting the number of available &sworker;s.
+    handle becomes too large without adjusting the number of available
+    Salt worker threads.
    </para>
    <para>
     For the correct value, refer to
--- a/xml/admin_administration.xml
+++ b/xml/admin_administration.xml
@@ -141,9 +141,9 @@ cmd.run "<replaceable>df -h</replaceable>"</command>
   <sect2 xml:id="sec.admin.salt.worker_threads">
    <title>Adjusting The Number Of Salt Worker Threads</title>
    <para>
-    It will sometimes be necessary to resize the cluster to adjust for workloads
-    or other factors. Salt will run into problems, if the number of nodes to
-    handle becomes too large without adjusting the number of available
+    It will sometimes be necessary to resize the &kube; cluster to adjust for
+    workloads or other factors. Salt will run into problems, if the number of
+    nodes to handle becomes too large without adjusting the number of available
     Salt worker threads.
    </para>
    <para>
--- a/xml/admin_administration.xml
+++ b/xml/admin_administration.xml
@@ -137,6 +137,52 @@ cmd.run "<replaceable>df -h</replaceable>"</command>
    Replace the example <command>df -h</command> with a command of your choice.
    The output will be produced in your current terminal session.
   </para>
+
+  <sect2 xml:id="sec.admin.salt.workers">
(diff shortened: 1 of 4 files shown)


xml/admin_administration.xml
```diff
@@ -137,6 +137,52 @@ cmd.run "<replaceable>df -h</replaceable>"</command>
    Replace the example <command>df -h</command> with a command of your choice.
    The output will be produced in your current terminal session.
   </para>
+
+  <sect2 xml:id="sec.admin.salt.workers">
+   <title>Adjusting The Number Of Salt Workers</title>
+   <para>
+    It will sometimes be necessary to resize the cluster to adjust for workloads
```
ereslibre: `resize the cluster` might indicate changing the number of masters/workers on the cluster. Maybe a rewording here would be better: `It will sometimes be necessary to increase the number of worker threads of the salt master`.

xml/deployment_installing_nodes.xml
```diff
@@ -993,15 +993,15 @@
...
      <para>
       If you quit out of the configuration process at this time by closing
       the browser window/tab, you will lose all input you made to the settings
       and have to restart the configuration from the beginning.
      </para>
      <para>
-      We recommend you leave the browser window/tab open while you perform the
-      installation of the other nodes in a new window.
+      You must leave the browser window/tab open while performing the
+      installation of the other nodes in a new window/tab.
```
ereslibre: I don't think the new or the old descriptions are accurate. On every step of the setup (when you click `Next`), we save that information in the database, and it's perfectly safe to close it and come to it afterwards. You will only lose whatever was on that page where you didn't click `Next`, but the previous steps data will survive across sessions.

xml/deployment_installing_nodes.xml
```diff
@@ -1082,6 +1093,17 @@
...
+      <title>Adjust Salt Cluster Size For More Than 40 Nodes</title>
+      <para>
+       If you are deploying a cluster with more than <literal>40</literal> overall
+       nodes, you must adjust the number of available &sworker;s before you
+       continue.
+      </para>
+      <para>
+       Refer to: <xref linkend="sec.deploy.requirements.system.cluster.salt_cluster_size"/>.
+      </para>
+     </important>
```
ereslibre: I might be losing some context from the diff, but why is this warning present twice?

xml/deployment_installing_nodes.xml
```diff
@@ -1184,9 +1206,19 @@
...
```
ereslibre: Three times?

xml/deployment_sysreqs.xml
```diff
@@ -47,6 +47,67 @@
    size. For the current maximum supported number of nodes, please refer to
    the Release Notes on <link xlink:href="https://www.suse.com/releasenotes/"/>.
   </para>
+
+  <sect2 xml:id="sec.deploy.requirements.system.cluster.salt_cluster_size">
+   <title>Salt Cluster Sizing</title>
+   <para>
+    &productname; relies on SaltStack (or Salt for short) to automate various
+    tasks. The actions are performed by so called "workers". A worker can handle
```
ereslibre: Hmmm... Salt's architecture in our product has the salt-master and salt-minions. The salt-minions are the agents that run different operations on all machines, and the salt-master has this notion of workers, that is the number of channels it has open in order to send commands and receive their result from the salt-minions.

I'm not sure if saying that "actions are performed by so called 'workers'" is accurate. Workers is just the concept salt-master has for how many concurrent channels can be open at any given time, that will connect to salt-minions.

xml/admin_administration.xml
r0ckarong: By "resizing the cluster" i mean adding nodes to the cluster that then make adjusting the worker threads necessary.

ereslibre: I see, that makes sense. Maybe a `Kubernetes cluster` in this case would be nice.

xml/deployment_installing_nodes.xml
ereslibre: Yes, they should be saved, but only when you cilck on `Next`.

xml/admin_administration.xml
r0ckarong: added

PR #98, opened 2018-10-16
Add Active Directory and openLDAP examples

r0ckarong:
Find changes here: https://w3.nue.suse.com/~mnapp/2018-10-16/book.caasp.admin/cha.admin.security.html#sec.admin.security.external.ldap.examples


lcavajani:
LGTM

PR #100, opened 2018-10-17
fix bsc#1046128

MaximilianMeister:
document usage of 'transactional-update grub.cfg'


r0ckarong:
I'll manually rebase this into develop after adding a note about the reboot for the grub config to take hold. Thanks for the fix!

PR #105, opened 2018-11-07
deployment: sysreqs: Fix typo when describing worker threads

hwoarang:
The worker threads are being assigned on the master node instead of
the minions. This also renames 'nodes' with 'minions' to be consistent
with the rest of the paragraph.


r0ckarong:
The whole deployment guide is under a major rework so I don't know if this error will be in the new version once it comes in. I will merge this, nonetheless.

Thanks for the contribution!

PR #108, opened 2018-11-12
docker command was missing

MaximilianMeister:


r0ckarong:
I rewrote a bunch of those for the simpler filter parameters. Must've done something wrong. I'll check the others again. Thanks!

PR #110, opened 2018-11-22
admin_administration: Fix note about single replica pods

hwoarang:
Even if a pod has a single replica, it's still possible to use the
'kubectl drain' command [1] to evict it from a node which is going
to be rebooted.

[1] https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/#use-kubectl-drain-to-remove-a-node-from-service


svenseeberg:
Thanks for the pull request @hwoarang.

PR #113, opened 2018-12-07
Update links to upstream k8s docs

r0ckarong:
CaaSP maintenance update will update k8s to 1.10.11 to migitgate CVE-2018-1002105
Update links to correct upstream docs version


r0ckarong:
Reworked and picked kubedoc entity manually

PR #125, opened 2019-01-16
Add notes about changed NTP config with cloud-init

### LINKED ISSUE #114: Assigning Roles to the Cluster Nodes with CaaSP v4
5.2.5.11  Assigning Roles to the Cluster Nodes

https://www.suse.com/documentation/suse-caasp-3/singlehtml/book_caasp_deployment/book_caasp_deployment.html#sec.deploy.cloud-init.user-data.caasp-roles

With CaaSP v4, the cluster role does no longer configure a NTP client. Instead the admin has to configure one (either chrony or systemd-timesyncd) with cloud-init himself. NTP syncronisaton across the cluster is required.

The following text needs to be adjusted:
"...and configure a timesync service with administration node as a reference. You do not have to install any NTP server, but if you need to use one, you need to disable the systemd-timesyncd first."

r0ckarong:
Fixes #114

### DIFF
--- a/xml/deployment_configuration.xml
+++ b/xml/deployment_configuration.xml
@@ -685,21 +685,24 @@ resolv_conf:
    <sect3 xml:id="sec.deploy.cloud-init.user-data.ntp-server">
     <title>NTP Server Configuration</title>
     <para>
-     You can also configure the NTP server.
+     You must configure at least one NTP server.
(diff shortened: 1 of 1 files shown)


r0ckarong:
Needs to be backported once the v4 ntp features are implemented.

PR #126, opened 2019-01-17
Fix docker command for obtaining container name

hwoarang:
Here is a couple of fixes related to obtaining the container name

The first one fixes a typo and the second one makes the docs consistent by using the same method to obtain the container name across all sections.


hwoarang:
@r0ckarong I forgot to say that both need backporting to v3

PR #128, opened 2019-01-25
Add troubleshooting for node acceptance

r0ckarong:
Fixes: https://bugzilla.suse.com/show_bug.cgi?id=1062351

https://w3.nue.suse.com/~mnapp/2019-01-25/book.caasp.admin/cha.admin.troubleshooting.html#sec.admin.troubleshooting.node_acceptance

### DIFF
--- a/xml/admin_troubleshooting.xml
+++ b/xml/admin_troubleshooting.xml
@@ -112,9 +112,9 @@
 Accepted Keys:
 admin
 ca
+231bb41553ac4102bb8fda8b2fd9d60d
 Denied Keys:
 Unaccepted Keys:
-231bb41553ac4102bb8fda8b2fd9d60d
 bbaf54654d5649f68a0af5f1914ecedf
 db79f30700974971b5e32d62ba0d5d76
 Rejected Keys:
@@ -143,7 +143,8 @@ Rejected Keys:
     </para>
 <screen>&prompt.user;<command>docker ps -q -f name=event-processor</command></screen>
     <para>
-     If it's not running, make sure the admin node is not under memory pressure.
+     If it's not running or has a significantly different age from the other
+     containers, make sure the admin node is not under memory pressure.
     </para>
    </listitem>
    <listitem>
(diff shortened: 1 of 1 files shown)


xml/admin_troubleshooting.xml
```diff
@@ -57,6 +57,123 @@
...
+<screen>&prompt.user;<command>docker exec -it $(docker ps --filter=name=salt-master -q) salt-key</command>
+Accepted Keys:
+admin
+ca
+Denied Keys:
+Unaccepted Keys:
+231bb41553ac4102bb8fda8b2fd9d60d
+bbaf54654d5649f68a0af5f1914ecedf
+db79f30700974971b5e32d62ba0d5d76
+Rejected Keys:
```
ereslibre: It's weird to provide this example right after we say `Check that the minion ID has been actually accepted`, in this example all keys are unaccepted except for `admin` and `ca` (accepted automatically)

xml/admin_troubleshooting.xml
```diff
@@ -57,6 +57,123 @@
...
+     accepting the node from &dashboard;.
+    </para>
+   </listitem>
+   <listitem>
+    <para>
+     Check that the event processor is running on the &admin_node;:
+    </para>
+<screen>&prompt.user;<command>docker ps -q -f name=event-processor</command></screen>
+    <para>
+     If it's not running, make sure the admin node is not under memory pressure.
```
ereslibre: Not only if it's not running, but also if it has a different age than the rest of the containers.

If the system is stable, all containers on the admin node should have the same age when you do `docker ps`. If some containers have different age that's probably because we were running out of memory and some processes were killed automatically, the kubelet restarted the containers and the age of the restarted ones is newer. This would indicate that something's going on on the admin node.

xml/admin_troubleshooting.xml
r0ckarong: Yup, makes sense I'll move one to the accepted line

xml/admin_troubleshooting.xml
r0ckarong: changed the wording to include this age difference

PR #132, opened 2019-01-29
Add note about cluster sizing reserves

### LINKED ISSUE #131: Please add a note that we should not saturate nodes
We should have a suggestion note that nodes should not be overloaded (or even committed) so workloads can have more chances from moving from one node to another during an update or maintenance window.

r0ckarong:
Closes #131

### DIFF
--- a/xml/deployment_sysreqs.xml
+++ b/xml/deployment_sysreqs.xml
@@ -37,9 +37,7 @@
     You should always size your cluster to accomodate your expected workloads and
     contingency resources for updates. During updates, worker nodes will become
     temporarily unavailable to process workloads and their assigned load must
-    be shifted somewhere else in the cluster. In very small clusters with only
-    few worker nodes this is sometimes not possible (e.g. two workers each with
-    51% utilization).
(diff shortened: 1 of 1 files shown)


xml/deployment_sysreqs.xml
```diff
@@ -31,6 +31,25 @@
...
   </para>
+  <important>
+   <title>Cluster Update Reserves</title>
+   <para>
+    You should always size your cluster to accomodate your expected workloads and
+    contingency resources for updates. During updates, worker nodes will become
+    temporarily unavailable to process workloads and their assigned load must
+    be shifted somewhere else in the cluster. In very small clusters with only
+    few worker nodes this is sometimes not possible (e.g. two workers each with
+    51% utilization).
```
ereslibre: Not sure about the example TBH, this case is assuming two workers of the same size and without specified can be misleading, since if you have two workers of very different sizes even if the biggest is at 20% can lead to trouble (depending on the difference between them). I would omit the example.

r0ckarong: ok will remove that example

PR #139, opened 2019-02-11
Fix salt commands for failed update recovery

### LINKED ISSUE #138: [doc] 3.1.4  Recovering From Failed Updates
3.1.4  Recovering From Failed Updates

https://www.suse.com/documentation/suse-caasp-3/singlehtml/book_caasp_admin/book_caasp_admin.html#sec.admin.software.transactional-updates.recovering

We should target only the real minions using

`salt -P "roles:(admin|kube-(master|minion))"`

otherwise we will target also the `ca` container.

r0ckarong:
Fixes #138


r0ckarong:
Cleaned up the format for this salt command across the rest of the docs. Thanks.

PR #146, opened 2019-02-15
Add Backup and Restore documentation

r0ckarong:
Preview: ~https://w3.nue.suse.com/~mnapp/2019-02-15/book.caasp.admin/cha.admin.backup.html~

Updated: https://w3.nue.suse.com/~mnapp/2019-03-01/book.caasp.admin/cha.admin.backup.html
PDF: https://w3.nue.suse.com/~mnapp/2019-03-01/book.caasp.admin_color_en.pdf

### DIFF
--- a/xml/admin_backup.xml
+++ b/xml/admin_backup.xml
@@ -200,7 +200,10 @@ velum_production | gzip > <replaceable>/var/tmp/db-backup-$(date +'%Y-%m-%d').sq
      <para>
       Run the following command to unpack the data backup:
      </para>
-<screen>&prompt.root;<command>tar xzf <replaceable>backup-admin-node-[date].tgz</replaceable> --exclude='mariadb*' --exclude='55-returner-credentials.conf' -C / 2> /var/tmp/log.txt</command></screen>
+<screen>&prompt.root;<command>tar xzf <replaceable>backup-admin-node-[date].tgz</replaceable> \
+--exclude='mariadb*' \
+--exclude='55-returner-credentials.conf' \
+-C / 2> /var/tmp/log.txt</command></screen>
      <para>
       Run the following command to load the database backup into the MariaDB
       container. Make sure to replace <literal>[date]</literal> with the actual
@@ -210,8 +213,9 @@ velum_production | gzip > <replaceable>/var/tmp/db-backup-$(date +'%Y-%m-%d').sq
 docker exec -i $(docker ps -qf name="velum-mariadb") \
 mysql --password=$(cat /var/lib/misc/infra-secrets/mariadb-root-password) velum_production</command></screen>
      <note>
+      <title>Cluster Member Hostnames Must Be Resolvable</title>
       <para>
-       Note: These instructions assume that all cluster member node hostnames
+       These instructions assume that all cluster member node hostnames
        are resolvable by one another via DNS.
       </para>
      </note>
--- a/xml/admin_backup.xml
+++ b/xml/admin_backup.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE chapter
+[
+  <!ENTITY % entities SYSTEM "entity-decl.ent">
+    %entities;
+]>
+<chapter version="5.1" xml:id="cha.admin.backup"
+  xmlns="http://docbook.org/ns/docbook"
+  xmlns:xi="http://www.w3.org/2001/XInclude"
+  xmlns:xlink="http://www.w3.org/1999/xlink">
+ <info>
+  <title>Disaster Recovery</title>
+  <dm:docmanager xmlns:dm="urn:x-suse:ns:docmanager">
+   <dm:bugtracker/>
+   <dm:translation>yes</dm:translation>
+  </dm:docmanager>
+ </info>
+
+ <sect1 xml:id="sec.admin.backup">
+ <title>Backup Procedure</title>
+ <para>
+
+ </para>
+ </sect1>
+
+ <sect1 xml:id="sec.admin.restore">
+ <title>Restore Procedure</title>
+ <para>
+
+ </para>
+ </sect1>
+</chapter>
--- a/xml/admin_backup.xml
+++ b/xml/admin_backup.xml
@@ -171,7 +171,9 @@ tar czf <replaceable>backup-admin-node-$( date +'%Y-%m-%d' ).tgz</replaceable> -
       <xref linkend="sec.deploy.nodes.admin_install" />.
      </para>
      <para>
-      There is no need to complete the configuration steps in &dashboard;.
+      You must be able to access &dashboard; indicating all containers are
+      correctly started. There is, however, no need to complete the
+      configuration steps in &dashboard;.
      </para>
     </step>
     <step>
--- a/xml/admin_backup.xml
+++ b/xml/admin_backup.xml
@@ -120,6 +120,7 @@
 tar czf <replaceable>backup-admin-node-$(date +'%Y-%m-%d').tgz</replaceable> -C / \
     var/lib/misc/salt-master-credentials \
     usr/share/salt \
+    etc/hosts \
     etc/pki usr/share/caasp-container-manifests \
     etc/salt/pki \
     etc/caasp \
--- a/xml/admin_backup.xml
+++ b/xml/admin_backup.xml
@@ -132,7 +132,8 @@ tar czf <replaceable>backup-admin-node-$(date +'%Y-%m-%d').tgz</replaceable> -C 
     var/lib/misc/ldap-config \
     var/lib/misc/ssh-public-key \
     etc/salt/pillar \
-    var/lib/misc/velum-secrets</command>
+    var/lib/misc/velum-secrets \
+    etc/machine-id</command>
  </screen>
     </step>
     <step>
@@ -199,7 +200,7 @@ velum_production | gzip > <replaceable>/var/tmp/db-backup-$(date +'%Y-%m-%d').sq
      <para>
       Run the following command to unpack the data backup:
      </para>
-<screen>&prompt.root;<command>tar xzf <replaceable>backup-admin-node-[date].tgz</replaceable> --exclude='mariadb*' -C / 2> /var/tmp/log.txt</command></screen>
+<screen>&prompt.root;<command>tar xzf <replaceable>backup-admin-node-[date].tgz</replaceable> --exclude='mariadb*' --exclude='55-returner-credentials.conf' -C / 2> /var/tmp/log.txt</command></screen>
      <para>
       Run the following command to load the database backup into the MariaDB
       container. Make sure to replace <literal>[date]</literal> with the actual
@@ -220,6 +221,12 @@ mysql --password=$(cat /var/lib/misc/infra-secrets/mariadb-root-password) velum_
       Once the files have been unpacked successfully, reboot the node.
      </para>
     <screen>&prompt.root;<command>reboot now</command></screen>
+    </step>
+    <step>
+     <para>
+      Once the new &admin_node; has restarted, ssh back onto it and run the following command.
+<screen>&prompt.root;<command> docker exec -it $(docker ps -qf name="salt-master") salt-key -A</command></screen>
+     </para>
     </step>
     <step>
      <para>
--- a/xml/admin_backup.xml
+++ b/xml/admin_backup.xml
@@ -16,17 +16,170 @@
   </dm:docmanager>
  </info>
 
+ <sect1 xml:id="sec.admin.assumptions">
+ <title>Assumptions</title>
+ <para>
+  A word of warning.
+ </para>
+ <screen>
+  (Marek Counts another thing not on this list..: this does not backup/restore persistent volumes/storage and they should be backuped/restored separately.)
+
+1) adding workers
+
+Force Remove, add new worker via Velum
+
+2) adding masters
+
+Force Remove, add new master via Velum
+
+3) we are not worried about etcd losing quorum (aka, losing multiple masters quickly)
+
+You will not be able to backup existing workloads
+
+4) we are worried about only v3 for this epic (not stated as far as I read)
+
+→ Needs task to "update for v4" whatever that means
+
+5) for admin we are worried about bare metal AND we assume the rest of the cluster is OK and running. We just need to start up a new Admin server, restore it and join it back to the cluster.
+
+If master quorum is broken, you're screwed
+
+6) etcd is etcd is etcd and backing up and restoring it is outside this epic.
+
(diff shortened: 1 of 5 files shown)


xml/admin_backup.xml
```diff
@@ -0,0 +1,394 @@
...
+     </para>
+    </step>
+    <step>
+     <para>
+     Verify the node is removed from the node list in Velum
+     </para>
+    </step>
+    <step>
+     <para>
+      Add a new master/worker node: <xref linkend="sec.deploy.nodes.worker_install" />
```
thehejik: The link here is not generated, it is only text formatted as italic font. The link above is working.

See https://w3.nue.suse.com/~mnapp/2019-02-15/book.caasp.admin/cha.admin.backup.html#sec.admin.backup.general.restore

thehejik: Another problem in this step is that the node(s) have to be upgraded to the same versions of packages as other node. Do we have somewhere described the procedure how to do manual upgrade of the node?

thehejik: Another problem in this step is that the node(s) have to be upgraded to the same versions of packages as the nodes in cluster (mainly salt-states, kubernetes). Do we have somewhere described the procedure how to do the upgrade of the node?

That part would be bit tricky especially if deployed from qcow2/vmdk image (but "SUSEConnect -r ; transactional-update dup" should solve it). On the other hand the update should be relatively simple when installing new node from ISO (either manually with registration+update or by using autoyast from velum).

xml/admin_backup.xml
```diff
@@ -0,0 +1,394 @@
...
+      SSH into the &admin_node;.
+     </para>
+    </step>
+    <step>
+     <para>
+      Change to the <filename>/var/tmp</filename> directory and package
+      important data into a tar file archive:
+     </para>
+<screen>&prompt.root;<command>
+tar czf <replaceable>backup-admin-node-$( date +'%Y-%m-%d' ).tgz</replaceable> -C / \
```
thehejik: produces correct "backup-admin-node-2019-02-26.tgz" filename but personally I would remove spaces close to the `date` command so it would look like this:

`tar czf <replaceable>backup-admin-node-$(date +'%Y-%m-%d').tgz</replaceable> -C / \`

xml/admin_backup.xml
```diff
@@ -0,0 +1,394 @@
...
+    var/lib/misc/velum-secrets</command>
+ </screen>
+    </step>
+    <step>
+     <para>
+      Then, export the MariaDB database and archive it:
+     </para>
+ <screen>&prompt.root;<command>docker exec $(docker ps -qf name="velum-mariadb") mysqldump \
+ --password=$(cat /var/lib/misc/infra-secrets/mariadb-root-password) \
+ velum_production | gzip > <replaceable>/var/tmp/db-backup-[$(date+'%Y-%m-%d')].sql.gz</replaceable></command>
```
thehejik: * Missing space between `date` and `+` sign ... date+ command not found
* Why the date output is enclosed in square brackets? It doens't look good and might be confusing
  - if you wanna really use square brackets please use them also for the `tar` command at Line120
* I would not start a newline with a leading space just bellow a line ending by `\` char (it produces double-spaces when copy&pasting)

This is working (without square brackets):

```
 <screen>&prompt.root;<command>docker exec $(docker ps -qf name="velum-mariadb") mysqldump \
--password=$(cat /var/lib/misc/infra-secrets/mariadb-root-password) \
velum_production | gzip > <replaceable>/var/tmp/db-backup-$(date +'%Y-%m-%d').sql.gz</replaceable></command>
 </screen>
```

produces "db-backup-2019-02-26.sql.gz"

xml/admin_backup.xml
```diff
@@ -0,0 +1,394 @@
...
+ </screen>
+    </step>
+    <step>
+     <para>
+      Finally, copy the created files <filename>backup-admin-node-[date].tgz</filename>
```
thehejik: The `tar` command for backup-admin-node-.tgz doesn't produce `-[2019-02-26].tgz` but `-2019-02-26.tgz` without square brackets.

xml/admin_backup.xml
```diff
@@ -0,0 +1,394 @@
...
+
+  <sect2 xml:id="sec.admin.backup.admin.restore">
+   <title>Restore Procedure</title>
+   <procedure>
+    <step>
+     <para>
+      Set up a new machine for the &admin_node; by following
+      <xref linkend="sec.deploy.nodes.admin_install" />.
+     </para>
+     <para>
```
thehejik: Please add a paragraph about the need to have registered and fully upgraded admin node here.

xml/admin_backup.xml
thehejik: What I don't like here is that the command doesn't normally return 0 because some changes happened in files during the archiving. It would be good idea to stop services/containers before performing a backup (or excluding all the dynamic files from the archive?).

The solution for this would be probably too complicated - something like `systemctl stop docker`; `tar czf ... `; `systemctl reboot` - to restart containers again because we need mariadb container up for mysqldump.

This is the issue with return code 1:
```
# tar czf backup-admin-node-$(date +'%Y-%m-%d').tgz -C /
...
tar: var/lib/mysql/ib_logfile1: file changed as we read it
tar: var/lib/mysql/ib_logfile0: file changed as we read it
tar: var/lib/misc/salt/cache/salt-master/master/tokens: file changed as we read it
admin:/var/tmp # echo $?
1

```

xml/admin_backup.xml
r0ckarong: No this is a quirk of Docbook and the way it handles cross references. It will `know` where the reference in the other file is but since it doesn't know where the URL is, it will spell out how to find it. There is no good way to fix this without hardcoding all links as of now. I will look into how this can be improved.

xml/admin_backup.xml
r0ckarong: - Fixed this and the instance above
- That's how Marek wrote it, I'm going to remove the brackets from the output but leave them in the examples to make clear where the date portion is.
- example was indented wrong in the source code, there is no leading space after the fix

xml/admin_backup.xml
r0ckarong: Currently we only describe how to upgrade all cluster nodes via Velum. This scenario is not covered.

xml/admin_backup.xml
```diff
@@ -0,0 +1,394 @@
...
+      SSH into the new &admin_node; machine and switch to the location where
+      you have stored the backup files.
+     </para>
+    </step>
+    <step>
+     <para>
+      Run the following command to unpack the data backup:
+     </para>
+<screen>&prompt.root;<command>tar xzf <replaceable>backup-admin-node-[date].tgz</replaceable> -C /</command></screen>
```
thehejik: This approach doesn't work on RO filesystem, first you have to switch fs to RW mode somehow (which is most likely not supported) or call `transactional-update shell` and do it there. I'm trying to find out working method.

```
admin # tar xzf backup-admin-node-2019-02-26.tgz -C /
tar: usr/share/salt/kubernetes/config/master.d/50-api.conf: Cannot open: File exists
tar: usr/share/salt/kubernetes/config/master.d/50-master.conf: Cannot open: File exists
tar: usr/share/salt/kubernetes/config/master.d/50-peer.conf: Cannot open: File exists
tar: usr/share/salt/kubernetes/config/master.d/50-reactor.conf: Cannot open: File exists
tar: usr/share/salt/kubernetes/config/master.d/50-returner.conf: Cannot open: File exists
tar: usr/share/salt/kubernetes/config/master.d: Cannot utime: Read-only file system
tar: usr/share/salt/kubernetes/config/master.d: Cannot change ownership to uid 0, gid 0: Read-only file system
tar: usr/share/salt/kubernetes/config: Cannot utime: Read-only file system
tar: usr/share/salt/kubernetes/config: Cannot change ownership to uid 0, gid 0: Read-only file system
```

xml/admin_backup.xml
```diff
@@ -0,0 +1,398 @@
...
+      important data into a tar file archive:
+     </para>
+<screen>&prompt.root;<command>
+tar czf <replaceable>backup-admin-node-$(date +'%Y-%m-%d').tgz</replaceable> -C / \
+    var/lib/misc/salt-master-credentials \
+    usr/share/salt \
+    etc/pki usr/share/caasp-container-manifests \
+    etc/salt/pki \
+    etc/caasp \
+    var/lib/misc/salt/cache \
```
thehejik: please add somewhere here `etc/hosts \` (without that the velum is not starting after restore)

thehejik:
The procedure for restoration of admin node is not working, so far I identified two problems:
* restoring of tgz file to RO filesystem doesn't work (use `transactional-update` shell env to extract)
* we don't backup /etc/hosts which is mandatory (without that file the velum wont start)
* when I pack also /etc/hosts file the velum is starting but the nodes are visible in "Nodes" and "Pending Nodes" at the same time with the same machine-id - I guess we would need to backup also salt-key data somehow.

xml/admin_backup.xml
thehejik: Ok, running `transactional-update shell` (t-u shell) and extracting the backup-admin.tgz file within the shell do the trick.

So please modify this step to have this meaning:
* move the file backup-admin.tgz to /root - because /var/tmp is not mounted in the t-u shell env
* run `transactional-update shell`
* `cd /root` and perform `tar zxf backup-admin.tgz -C /`
* type `exit` to leave t-u shell env
* call `systemctl reboot` to take a new btrfs snapshot with restored data into account

Continue with db restore - `t-u shell` is not needed for that...

xml/admin_backup.xml
r0ckarong: Done

xml/admin_backup.xml
Klaven: this is not needed `/var/tmp` should be writable and is probably why it was not avalible within the transactional-update shell

xml/admin_backup.xml
Klaven: So, the goal is to NOT shut down the server each time you backup (nightly) the only current issue is `/var/lib/mysql` and I am working on an update too this one. will update this one shortly. All the other ones should be fine UNLESS they are running an orchestration and this should be noted that they should not do during a backup.

xml/admin_backup.xml
thehejik: @Klaven sorry, but I'm not getting your point - we need to switch to `t-u shell` to do `tar xz ...` because of RO fs but in this shell `/var/tmp/` is not mounted so we need to move the backup to /root (or another location visible under `t-u shell`)

thehejik: We clarified with @Klaven that `t-u shell` is really not needed...  As we restore only into writable dirs like /etc/ and the rest is not interesting for the backup, but anyway the command will produce lot of garbage output/errors

@r0ckarong you can ignore this thread

xml/admin_backup.xml
```diff
@@ -0,0 +1,398 @@
...
+      </para>
+     </note>
+    </step>
+    <step>
+     <para>
+      Once the files have been unpacked successfully, reboot the node.
+     </para>
+    <screen>&prompt.root;<command>reboot now</command></screen>
+    </step>
+    <step>
```
thehejik: One more step is needed - users have to re-accept salt minions (either in Velum or by SSH) because we don't do a backup of the salt-master container so we are loosing info about previously accepted salt minions.

Without this step the velum will show up all the nodes under "Nodes" and "Pending Nodes" at the same time with the same machine-id...

Run this command over SSH to re-accept the salt minions before entering Velum and then the issue will be not visible for users (and user impression will be better):
docker exec -i $(docker ps -qf name="salt-master") salt-key --accept-all --yes

thehejik: It appears that we do backup of etc/salt/pki which should contain all salt-key minion data. I will re-validate if this step is really needed once again later.

thehejik: Unfortunately this problem persists.
reaccept_minions_after_restore
(Ignore wrong node hostnames, it's known issue - it was there before I performed b&r)

I repeated all the backup & restore steps again by following updated doc https://w3.nue.suse.com/~mnapp/2019-03-01/book.caasp.admin/cha.admin.backup.html#sec.admin.backup

xml/admin_backup.xml
```diff
@@ -0,0 +1,398 @@
...
+     <para>
+      SSH into the new &admin_node; machine and switch to the location where
+      you have stored the backup files.
+     </para>
+    </step>
+    <step>
+     <para>
+      Run the following command to unpack the data backup:
+     </para>
+<screen>&prompt.root;<command>tar xzf <replaceable>backup-admin-node-[date].tgz</replaceable> --exclude='mariadb*' -C / 2> /var/tmp/log.txt</command></screen>
```
thehejik: @Klaven Why do you exclude mariadb*? I'm asking because it still produces error code 2 and log is full of "ro filesystem" errors anyway...

xml/admin_backup.xml
thehejik: I found maybe related problem - admin node changes its `/etc/machine-id` and possibly it may lead to some unpredictable problems (like this???)

Good thing is that salt-master machine-id didn't change and admin node is named 'admin' for salt (minion name) and not `<machine-id>`. In addition entry in /etc/hosts for admin also doesn't contain `<machine-id>` hash.

Anyway is there any way how can we preserve admin's machine-id? What about to add `/etc/machine-id` into backup?

**original node**
```
admin:~ # cat /etc/machine-id 
02a0cffb2fa54152b458ede6a551796c
admin:~ # docker exec -i $(docker ps -f name=salt-master -q) cat /etc/machine-id
7b9462cb98a677069326e3f05b2926d5
```

**restored node**
```
admin:~ # cat /etc/machine-id 
11a0cffb2fa54152b458beefa551deed
admin:~ # docker exec -i $(docker ps -f name=salt-master -q) cat /etc/machine-id
7b9462cb98a677069326e3f05b2926d5
```

Klaven: Let me give it a try.

xml/admin_backup.xml
Klaven: The reason being that mariadb has already generated users and passwords and those passwords are stored in the files i'm excluding. We just want to keep the new generated password files as everything uses those files to talk to mariadb.

xml/admin_backup.xml
thehejik: @Klaven any progress here?

thehejik: ah, now I see where is the problem with re-accepting minions. When you deploy a new admin node a current worker nodes are trying contact their salt-master (not restored yet) on the same address as it was before but the admin doesn't know them yet... Anyway --accept-all is ok for me.

PR #154, opened 2019-02-28
cilium: Add more detailed description and example

mrostecki:
Provice more detailed description of Cilium (what it is, what
technologies does it use etc.) and an example of checking whether it
works properly after deployment.


jordimassaguerpla:
If @r0ckarong is ok, I am.

PR #191, opened 2019-04-10
Add note about rebooting masters

dannysauer:
Tiny change for a large discussion. :D 

Here's a proposed addition adding an extra step to reboot the master nodes after applying the new CA certificates.  Feel free to change the wording completely or move it around the document. :)

### DIFF
--- a/xml/admin_security.xml
+++ b/xml/admin_security.xml
@@ -1001,6 +1001,14 @@ roleRef:
        user must apply this update by clicking the "Apply changes" button at
        top-right.
       </para>
+      <para>
+       After the change has been applied to all nodes, the master nodes must be
+       rebooted one at a time in order for all daemons to recognize the new CA
+       certificates.  Failure to complete this step will result in failure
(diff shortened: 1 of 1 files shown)


r0ckarong:
Could you enlighten me which discussion you are referring to? The wording itself sounds right to me but I'm missing the context to understand which problem it solves and if we need to change something else.

dannysauer:
@r0ckarong This may not need to be done, given the direction shift doc'd in issue SUSE/caasp-playground#171.  It looks like we're going to try to make the service restarts / reboot automatic with Salt after all, potentially rendering this doc update unnecessary. :)  I'll update the PR in a day or two based on the outcome of resolving that issue.

r0ckarong:
@dannysauer I'll put it on hold. If this resolves itself just close the PR.

dannysauer:
Sorry; I could've sworn I closed this as unnecessary before.  Closing it now. :)

PR #196, opened 2019-04-15
Errata for VMware

thehejik:
I identified few issues in documentation for vNext on VMware.

@r0ckarong please check.


r0ckarong:
Thanks!

PR #197, opened 2019-04-15
Provide more information about ssh-agent

flavio:
Give some basic instruction about how to interact with ssh-agent.

Can help when someone doesn't know how to interact with `ssh-agent`, see https://github.com/SUSE/caaspctl/pull/78


dannysauer:
Some of the claims in this documentation are inaccurate. I'm adding this comment here so hopefully I'll remember to correct them when I get home. :)

PR #203, opened 2019-05-02
Added a note about node registration do the applying updates section

nkoranova:


xml/admin_software.xml
```diff
@@ -279,6 +279,15 @@
 
   <sect2 xml:id="sec.admin.software.transactional-updates.installation">
    <title>Applying Updates</title>
+   <note>
+    <title>Registering nodes</title>
+    <para>
+     In order to update nodes, you should first register them at SUSE Customer
```
r0ckarong: It should say "you must first register..." because without registration it simply won't work. "Should" implies that it's possible.

xml/admin_software.xml
```diff
@@ -279,6 +279,15 @@
 
   <sect2 xml:id="sec.admin.software.transactional-updates.installation">
    <title>Applying Updates</title>
+   <note>
+    <title>Registering nodes</title>
+    <para>
+     In order to update nodes, you should first register them at SUSE Customer 
+     Center, SMT or RMT. If you did not register during the installation, see 
+     <link xlink:href="https://www.suse.com/documentation/suse-caasp-3/singlehtml/book_caasp_deployment/book_caasp_deployment.html#sec.configuration.suseconnect"/>
```
r0ckarong: `link` is only for external links. Outside websites and such.

`xref` is used when you want to point at a section of the same document set (a set is what is configured together in the `MAIN.caasp.xml` file (https://github.com/SUSE/doc-caasp/blob/maintenance/CaaS3/xml/MAIN.caasp.xml). 

You would replace this link with

`<xref linkend="sec.configuration.suseconnect"/>`

`sec.configuration.suseconnect` is a unique XML ID that identifies a certain section or element across the documents. If the external link changes it will break. With the xref element the link is build dynamically and will always work as long as the documents are published in the same location.

Read more about it here: http://docserv.nue.suse.com/documents/Style_Guide/docu_styleguide/html/#sec.cross_reference

xml/admin_software.xml
```diff
@@ -279,6 +279,15 @@
...
+     &mdash; the SUSE CaaS Platform 3 Deployment Guide.
```
r0ckarong: Please replace the product names in this paragraph with the respective entities. For example instead of writing out `SUSE Customer Center` you should use `&scc;`. The entities are defined here: 

https://github.com/SUSE/doc-caasp/blob/maintenance/CaaS3/xml/entity-decl.ent#L25

most importantly instead of having to write out all of `SUSE CaaS Platform 3` just use `&productname; &productnumber;`. This also makes it a lot easier to maintain versions because you just need to change the version number in the entities file and not across X locations around all files.

Please check out the Entities section in the Hitchhikers Guide (http://docserv.nue.suse.com/documents/Style_Guide/docu_styleguide/html/#sec.entity) and look through the entities file to get familiar with what you can replace.

xml/admin_software.xml
nkoranova: Aha, thanks! I used the external link tag because it refers to a different document (the Deployment guide) and I assumed you can only use xref for references inside one document. I just found a place, where this issue is dealt with in yet another way: https://www.suse.com/documentation/suse-caasp-3/singlehtml/book_caasp_admin/book_caasp_admin.html#sec.admin.troubleshooting.proxy
Do we want to do something about that?

r0ckarong: I think we can leave the section in troubleshooting like that. It is just specifically meant to forward people who have an auth proxy between their cluster and scc to find the right section. Since the configuration document explains all the rest in detail there is no need for more links.

PR #204, opened 2019-05-03
Architecture description

### LINKED ISSUE #199: Create skeleton for architecture description
Use inputs already known from:

- https://confluence.suse.de/display/SUSECaaSPlatform4/Architectural+Overview?src=contextnavpagetreemode
- https://confluence.suse.de/download/attachments/217350262/CaaSP%20evolution.pdf?version=1&modificationDate=1552900311000&api=v2
- https://confluence.suse.de/display/SUSECaaSPlatform4/Summary?src=contextnavpagetreemode
- https://confluence.suse.de/pages/viewpage.action?pageId=231211033

EDIT:
Content for now lives here: https://github.com/SUSE/doc-caasp/blob/architecture-description/adoc/architecture-description.adoc

r0ckarong:
[Architecture Description](https://github.com/SUSE/doc-caasp/blob/architecture-description/adoc/architecture-description.adoc)

[Include File Architecture Updates](https://github.com/SUSE/doc-caasp/blob/architecture-description/adoc/architecture-updates.adoc)


adoc/architecture-description.adoc
```diff
@@ -0,0 +1,316 @@
...
+
+You can deploy {productname} onto physical servers or use it on virtual machines. After deployment, it is immediately ready to run and provides a highly-scalable cluster.
+
+{productname} inherits benefits of SUSE Linux Enterprise and uses tools and technologies well-known to system administrators such as cloud-init, AutoYaST and Terraform.
+
+{productname} uses the Btrfs file system with the following characteristics:
+
+* The root file system and its snapshots are read-only.
+* Sub-volumes for data sharing are read-write.
+* {productname} introduces overlays of the /etc directories used by cloud-init and Salt.
```
ereslibre: This whole section can be removed, SLE15 SP1 uses `xfs`.

adoc/architecture-description.adoc
```diff
@@ -0,0 +1,316 @@
...
+|===
+
+
+== Supported Platforms
+
+* SUSE OpenStack Cloud 8
+* VMWare ESXi 6.0.0 Update 3
+* Bare Metal
+** KVM
+** x86_64
```
ereslibre: I think it's a little weird to mix architectures with platforms here. It would be good to have a clear statement on what architectures are supported on every platform. If only `x86_64` is supported right now I think something along the lines of:

```
== Supported platforms

* SUSE OpenStack Cloud 8
* VMWare ESXi 6.0.0 Update 3
* Bare Metal
```

```
== Supported architectures

* x86_64
```

would be better. Also, KVM is something different than bare metal. Bare metal would imply no hypervisor at all, whereas KVM is a virtualization platform. I think for the time being KVM can be out of the list.

adoc/architecture-description.adoc
```diff
@@ -0,0 +1,316 @@
...
+
+Any other SLE 15 SP1 supported platform.
```
ereslibre: This is not necessarily true, we state explicitly what platforms and architectures we support.

adoc/architecture-description.adoc
```diff
@@ -0,0 +1,316 @@
...
+
+=== Public Cloud
+
+* Google Kubernetes Engine (GKE)
```
ereslibre:
```suggestion
* Google Compute Engine (GCE)
```

since `GKE` is a managed Kubernetes solution from Google :)

adoc/architecture-updates.adoc
```diff
@@ -0,0 +1,146 @@
+=== Upgrades of OS components (automated)
+
+By default {productname} clusters automatically apply all the patches that are marked as non
+interactive. These patches are safe to be applied since they should not cause any side effect.
+
+However, some patches need the nodes to be rebooted in order to be activated. This is the
+case for example of kernel updates or some packages updates (like glibc).
```
ereslibre:
```suggestion
case for example of kernel updates or some package updates (like glibc).
```

adoc/architecture-updates.adoc
```diff
@@ -0,0 +1,146 @@
...
+* Open a standard kubernetes UI (eg: ​kubernetes dashboard​).
+* Click on the node.
+* Look at the annotations associated to the node.
+
+Text mode:
+
+* Go to a machine with a working kubectl (meaning I can connect to the cluster).
+* Ensure I’ve the caasp kubectl plugin installed. This is a simple statically linked binary
+that SUSE distributes as RPM.
+* Execute the `kubectl caasp nodes status` command.
```
ereslibre:
```suggestion
* Execute the `kubectl caasp cluster status` command.
```

adoc/architecture-updates.adoc
```diff
@@ -0,0 +1,146 @@
...
+
+System administrators can also fine tune the kured deployment to prevent its agent from
+rebooting machines where special workloads are running. For example: I can prevent kured
+from rebooting nodes running computational workloads until their pods are done. To achieve
+that I just have to instruct all my users to add special labels to the pods they don’t want to see
+interrupted due to the node being rebooted.
+
+==== Interactive upgrades
+
+Some updates might cause damages to the running workloads. For example, docker updates
```
ereslibre: Not sure if we should keep this example, as CaaSP v4 is now exclusively crio based.

adoc/architecture-updates.adoc
```diff
@@ -0,0 +1,146 @@
...
+can cause all the containers running on a machine to be suddenly killed.
+Note well:​ these interactive updates are currently being referenced by this document as
+“disruptive upgrades”.
+
+System administrators don’t have to worry about disruptive upgrades. {productname} will
+automatically apply them making sure no disruption is caused to the cluster.
+That happens because nodes with disruptive upgrades are updated one at a time, a bit like
+when nodes are automatically rebooted. Moreover {productname} will take care of cordoning and
```
ereslibre:
```suggestion
when nodes are automatically rebooted. Moreover {productname} will take care of draining and
```

ereslibre: The wording is a little off: `Moreover {productname} will take care of draining the node before these updates are applied, and uncordoning it afterwards`. The uncordoning should happen only after the machine is back to a running state and with latest updates applied.

adoc/architecture-updates.adoc
```diff
@@ -0,0 +1,146 @@
...
+* Are the disruptive upgrades being applied?
+
+=== Upgrades of OS components (not automated)
+
+As a systems administrator I want my cluster to be updated only when I consider it’s necessary,
+with human intervention as a must for every package update, so that I’m in complete control of
+when my cluster is updated.
+By default {productname} clusters have some updates applied automatically. Nodes can also be
+rebooted in an automatic fashion under some circumstances.
+To prevent that from happening I have to disable a systemd timer on each node of the cluster.
```
ereslibre: It's also possible to annotate nodes so kured won't automatically reboot the node (even leaving the systemd timer running), see: https://github.com/weaveworks/kured#disabling-reboots

This has the advantage that a user that wants to temporarily or permanentely disable reboots don't need SSH access to the machine, they can annotate the node using `kubectl` if their RBAC policies allow the user to do so.

adoc/architecture-description.adoc
r0ckarong: >  It would be good to have a clear statement on what architectures are supported on every platform.

Fully agree!

adoc/architecture-updates.adoc
r0ckarong: I took the text from the "CaaSP evolution vNext" document. If this has changes please adjust it.

PR #206, opened 2019-05-06
Describe Default PodSecurityPolicies

### LINKED ISSUE #184: Describe security policy handling in v4
https://github.com/SUSE/avant-garde/issues/15

r0ckarong:
Closes #184 

Rendered page: https://github.com/SUSE/doc-caasp/blob/describe-psp/adoc/admin-security-psp.adoc


r0ckarong:
> LGTM. Great page!
> 
> Only one thing that concerns me is that you have to keep in mind to update the PSP content (since you have pasted them into the doc) every time the PSP manifests changes as well.

That's why I called them "examples" I can remove them if that makes more sense. Do we expect people to copy+paste those examples and try to deploy them manually?

drpaneas:
> That's why I called them "examples". I can remove them if that makes more sense. Do we expect people to copy+paste those examples and try to deploy them manually?

No we don't expect them to change them. They can add their own, but by no means change those.

r0ckarong:
> > That's why I called them "examples". I can remove them if that makes more sense. Do we expect people to copy+paste those examples and try to deploy them manually?
> 
> No we don't expect them to change them. They can add their own, but by no means change those.

OK so I will remove the privileged file and leave only the unprivileged as an example.

ereslibre:
A small comment, we might add more PSP's in the future being privileged in different senses (not as widely as the privileged one), but they would be created by us just as "recipes". How the user would bind them (or even if they will bind them and to what) would be entirely up to the user.

r0ckarong:
> A small comment, we might add more PSP's in the future being privileged in different senses (not as widely as the privileged one), but they would be created by us just as "recipes". How the user would bind them (or even if they will bind them and to what) would be entirely up to the user.

@ereslibre How would you phrase that? I am trying to convey that configuring roles and bindings are up to the user. How would this work as a  "recipe"? Will we provide examples for specific configurations and then the user modifies them and binds them where they want them?

ereslibre:
@r0ckarong not sure if it's worth mentioning now since we only have privileged and unprivileged.

Think of PSP's as a "set of permissions". We provide different PSP's as different recipes in that each one has a different "set of permissions". For example:

- Privileged: allows absolutely everything.
- Unprivileged: does not allow anything.
- Network privileged (doesn't exist, example): Unprivileged + privileged in networking terms.

"How" the user binds these PSP's to their running workloads is a matter of the user, and depends a lot on how they structure their workloads and what their workloads needs are. Whether the user decides to use our "Network privileged" PSP, how and where is entirely up to them; in case it existed, we are just providing them a useful default "set of permissions" where pods would run unprivileged in general, but with access to the host networking.

In our case, right now, we are binding the `kube-system` namespace default service account to the privileged PSP, because user workloads should not run on this namespace.

r0ckarong:
> @r0ckarong not sure if it's worth mentioning now since we only have privileged and unprivileged.
> 
> Think of PSP's as a "set of permissions". We provide different PSP's as different recipes in that each one has a different "set of permissions". For example:
> 
>     * Privileged: allows absolutely everything.
> 
>     * Unprivileged: does not allow anything.
> 
>     * Network privileged (doesn't exist, example): Unprivileged + privileged in networking terms.
> 
> 
> "How" the user binds these PSP's to their running workloads is a matter of the user, and depends a lot on how they structure their workloads and what their workloads needs are. Whether the user decides to use our "Network privileged" PSP, how and where is entirely up to them; in case it existed, we are just providing them a useful default "set of permissions" where pods would run unprivileged in general, but with access to the host networking.
> 
> In our case, right now, we are binding the `kube-system` namespace default service account to the privileged PSP, because user workloads should not run on this namespace.

I did understand what you meant but I was trying to figure out if there is some reasonable way to mention this at this stage. In the long run we need to explain how to use RBAC/Role/RoleBinding and PSP to actually secure a workload as desired. We've got some pretty painful feedback from our sales people about how much of a burden this process is to them and the customers.

I think once we have more of these "recipes" we can try to explain this further. Right now I don't see any other way than diving in head first and explaining everything and I don't know if we have time for that (we don't have the resources at least).

PR #207, opened 2019-05-07
fixes for previous commit

### LINKED ISSUE #145: Add info about flexvolume drivers
https://github.com/kubic-project/salt/pull/724#issuecomment-463628747

nkoranova:
Properly Closes #145

### DIFF
--- a/xml/admin_misc.xml
+++ b/xml/admin_misc.xml
@@ -119,8 +119,8 @@ cmd.run "ln -sf <replaceable>/usr/share/zoneinfo/Europe/Berlin</replaceable> /et
    the flexvolume directory path, which is usually stored as an environment 
    variable, e.g. for rook:
   </para>
-  <screen>
-   &prompt;<command>FLEXVOLUME_DIR_PATH=/usr/libexec/kubernetes/kubelet-plugins/volume/exec/</command>
+<screen>
+<command>FLEXVOLUME_DIR_PATH=/usr/libexec/kubernetes/kubelet-plugins/volume/exec/</command>
   </screen>
   <para>
    For more information on DaemonSets, refer to <xref linkend="cha.user.daemonset"/> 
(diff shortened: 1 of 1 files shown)


xml/admin_misc.xml
```diff
@@ -119,8 +119,8 @@ cmd.run "ln -sf <replaceable>/usr/share/zoneinfo/Europe/Berlin</replaceable> /et
    the flexvolume directory path, which is usually stored as an environment 
    variable, e.g. for rook:
   </para>
-  <screen>
-   &prompt;<command>FLEXVOLUME_DIR_PATH=/usr/libexec/kubernetes/kubelet-plugins/volume/exec/</command>
+<screen>
+<command>FLEXVOLUME_DIR_PATH=/usr/libexec/kubernetes/kubelet-plugins/volume/exec/</command>
```
r0ckarong: Stupid question: Why is this a `<command>` tag?

nkoranova: That's a valid question. Because I used the formatting that was above in the document and then only removed the prompt part. 
But coincidentally because it is longer than 30 chars, so it should be a` <screen> ` and not just a ` <command>` and because the guide says to "Use markup for commands even inside <screen/> environments." http://docserv.nue.suse.com/documents/Style_Guide/docu_styleguide/html/#sec.command_line
Btw should this actually be a `<varname/>`?

r0ckarong: Yes, please change to `<varname>`.

nkoranova: > Yes, please change to `<varname>`.

ok, that's what you were hinting at.

r0ckarong: I was just wondering why it should have been a command. If the same line had been

```
export FLEXVOLUME_DIR_PATH=...
```

then it would have been a command. That can happen easily but for me I can't distinguish if it was a mistake in the command or the tag.

nkoranova: Yeah, I know what you mean. I'm glad you pointed it out, thanks. :) I didn't realize that you could use a variable tag, so I thought command was good enough.

PR #217, opened 2019-05-14
Add Nutanix Acropolis to the Supported Environments chapter

karras:
Currently the CaaSP "Supported Environments" section does not cover Nutanix Acropolis as a virtualization infrastructure. Acropolis itself is either using its own KVM-based hypervisor AHV or another vendor like VMWare (which are already listed as supported). From a technical point of view it should not make much difference running CaaSP workloads on there as any other virtualization layer.

Can you please check back internally and get the green light to list Nutanix as an officially supported environment as well? I had some discussions with Andreas Jäger about that, he should know the details in case of doubts. ;)

### DIFF
--- a/xml/deployment_sysreqs.xml
+++ b/xml/deployment_sysreqs.xml
@@ -449,6 +449,11 @@
        Oracle VM 3.3
       </para>
      </listitem>
+     <listitem>
+      <para>
+       Nutanix Acropolis
+      </para>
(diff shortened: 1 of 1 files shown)


r0ckarong:
@ajaeger This is only relevant for v3 right?

ajaeger:
Yes, this is for v3.

PR #222, opened 2019-05-20
Bare metal deployment guide

Klaven:
@r0ckarong I'm struggling here. for some reason the references are not working. Also, I refereed to it as `skuba`, not sure if I should be doing that or `caaspctl`. I feel like the one section about where/how to get the autoyast file should go away and we should just link the example when the code is open sourced.


r0ckarong:
@Klaven could you please allow maintainer edits for this pull request: 
https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

This makes work a lot easier. Thanks

flavio:
I looks like @dcoopr has already spotted this issue and proposed the same remediation I was suggesting: https://bugzilla.suse.com/show_bug.cgi?id=1135998

adoc/deployment-bare-metal.adoc
```diff
@@ -0,0 +1,67 @@
...
+[IMPORTANT]
+====
+These instructions currently do not describe how to set up a load balancer.
+This will be added in future versions. You must provide your own load balancing
+solution that directs access to the master nodes.
+====
+
+[NOTE]
+====
+The autoyast file generated by skuba is a template. It has the base requirements.
```
flavio: Is the skuba already capable of generating this template?

Klaven: no. wording updated.

jordimassaguerpla:
You may need to review the docs based on the result of this https://github.com/SUSE/skuba/issues/279

Klaven:
@jordimassaguerpla the person who completes #279 should update the documentation once it is complete. nothing to hold this PR for.

adoc/deployment-bare-metal.adoc
```diff
@@ -0,0 +1,73 @@
+include::entities.adoc[]
+:isofile: SLE-15-SP1-Installer-DVD-x86_64-Build225.1-Media1.iso
+:isolink: http://download.suse.de/install/SLE-15-SP1-Installer-GMC1/
```
thehejik: we shouldn't refer to internal server, anyway GMC1 is outdated now, current is http://download.suse.de/install/SLE-15-SP1-Installer-GMC3/SLE-15-SP1-Installer-DVD-x86_64-GMC3-DVD1.iso

adoc/deployment-bare-metal.adoc
```diff
@@ -0,0 +1,73 @@
...
+
+== Deployment on VMWare
```
thehejik: Is this related to vmware or baremetal?

adoc/deployment-bare-metal.adoc
jordimassaguerpla: How should we do it? Remove the iso from autoyast and update docs? Or how should we do it?

jordimassaguerpla: The official DVD can be download from https://www.suse.com/betaprogram/sle-beta/

adoc/deployment-bare-metal.adoc
r0ckarong: I created those for VMWare. If we have links for baremetal I would take them too. I want to be able to have links to the media that don't change much.

adoc/deployment-bare-metal.adoc
r0ckarong: Those are outdated values for the attributes. If you give me new URLs I can use those. The places that actually refer to the downloads already point to the sle beta page.

adoc/deployment-bare-metal.adoc
thehejik: i think it was resolved already by Klaven

jordimassaguerpla:
see https://github.com/Klaven/doc-caasp/pull/1 for fixing https://github.com/SUSE/avant-garde/issues/389

r0ckarong:
@Klaven I'm missing several things in the current version:

1. The AutoYaST example file sets up only a node but not a master. How does one set up a master? Which pattern needs to be changed?
2. Will this actually work like this? The "install required packages" comes before the "register the repositories holding the package" in the file. Is this parsed out of order?
3. How do you define the size of the cluster with this method?
4. Do you simply repeat these steps for as many nodes as you need?

thehejik:
I'm involved in autoyast as well so I will try to answer...

> @Klaven I'm missing several things in the current version:
> 
>     1. The AutoYaST example file sets up only a node but not a master. How does one set up a master? Which pattern needs to be changed?
There is no difference, master and worker nodes are using the very same pattern

>     2. Will this actually work like this? The "install required packages" comes before the "register the repositories holding the package" in the file. Is this parsed out of order?
If you are talking about item order in autoyast.xml file it doesn't matter - SLE will handle the item order automatically during installation.

> 
>     3. How do you define the size of the cluster with this method?
>     4. Do you simply repeat these steps for as many nodes as you need?

yup, just repeat these steps for as many nodes you need (or how many machines you have)

r0ckarong:
@thehejik @Klaven Please review the updated version.

adoc/deployment-bare-metal.adoc
```diff
@@ -28,39 +28,75 @@ Refer to the official {ay} documentation for more information: link:https://www.
...
++
+. Host the {ay} files on a webserver reachable inside the network your are installing the cluster in.
+
+=== Provisioning The Cluster Nodes
+
+Once the {ay} file is available in the network of the cluster you can start deploying machines.
+
+The default production scenario consists of 8 nodes:
+
+* 2 Load Balancers
```
thehejik: I don't think that `skuba` command supports more than one load-balancer.

adoc/deployment-bare-metal.adoc
```diff
@@ -28,39 +28,75 @@ Refer to the official {ay} documentation for more information: link:https://www.
...
+* 3 Workers
+
+Depending on the type of load balancer you wish to use, you need to deploy at least 6 machines to serve as cluster nodes and provide 2 load balancers from the environment.
+
+[TIP]
+If you do not wish to use infrastructure load balancers, please deploy additional machines and refer to <<loadbalancer>>.
+
+==== Setting Up The Master Nodes
+
+FIXME: What are the changes needed for the {ay} file to make this work as a master?
```
thehejik: you can delete this, no changes are needed

adoc/deployment-bare-metal.adoc
```diff
@@ -0,0 +1,100 @@
...
+
+[NOTE]
+====
+The {ay} file found in `skuba` is a template. It has the base requirements.
+This {ay} file should act as a guide and should be updated with your companies standards.
+====
+
+[NOTE]
+====
+To account for hardware/platform specific setup criteria (legacy BIOS vs. (U)EFI, drive partitioning, networking etc.),
```
thehejik: I've verified that current ay file is compatible with BIOS and EFI as well.

adoc/deployment-bare-metal.adoc
r0ckarong: This is how it's described in the architecture description. Any idea how this would be handled?

thehejik: No idea, maybe it can be achievable by using round-robin DNS entry (two lb hosts with the same public IP) but I would describe only a case with 1 LB. After clarifying we can modify again.

r0ckarong: I talked to Rafa we need to clarify the load balancer stuff more.

thehejik:
Apart of the unclear LBs count LGTM

adoc/deployment-bare-metal.adoc
thehejik: Now we refer to official SLES deployment guide so this is not needed anymore.

r0ckarong:
The LB topic will be clarified at a later date with: https://github.com/SUSE/doc-caasp/issues/261.
This PR will be merged.

PR #235, opened 2019-05-24
Add example nginx load balancer (Closes #160)

### LINKED ISSUE #160: Add more information about loadbalancer configuration
https://bugzilla.suse.com/show_bug.cgi?id=1124651

> if the customer is deploying CaaSP in public cloud then you can deploy using the loadbalancer type. If he deploys in BareMetal he has to provide an external load balancer

r0ckarong:
Closes https://github.com/SUSE/doc-caasp/issues/160

### DIFF
--- a/adoc/deployment-bootstrap.adoc
+++ b/adoc/deployment-bootstrap.adoc
@@ -1,3 +1,4 @@
+[[bootstrap]]
 == Bootstrapping the cluster
 
 === Cluster Deployment
--- a/adoc/deployment-loadbalancer.adoc
+++ b/adoc/deployment-loadbalancer.adoc
@@ -14,6 +14,8 @@ a failed request. The original request is lost and not forwarded to an available
 alternative server.
 
 This load balancer configuration is therefore only suitable for testing and proof-of-concept (POC) environments.
+
+For production environments we recommend the use of link:https://www.suse.com/documentation/sle-ha-15/index.html[{sle} {hasi} 15]
 ====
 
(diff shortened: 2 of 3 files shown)


r0ckarong:
@lcavajani 
what is the right time to set up the load balancer if you have to do this manually? before deploying the rest of the cluster or afterwards? 

You have to provide the loadbalancer during the bootstrap process instead of the first master right? so chicken egg problem a bit. How can I configure the loadbalancer with the node IP/FQDN if those don't exist but they only exist after I use the loadbalancer to bootstrap the cluster?

First manually bootstrap the nodes, then configure loadbalancer?

r0ckarong:
From Rafa:
> Ideally you should set up the load balancer before anything else, even if you have only one master (so it'll be easier to grow later if you plan to add HA afterwards). The loadbalancer can be set up separately before the first init. IP addresses can be reserved for the nodes that will be masters, so you let them aside, configure the loadbalancer pointing to those and init the first master and join the rest

r0ckarong:
@kkaempf @flavio Could you please provide me with the official stance on how load balancer is going to be supported and what we need to explain to people.

lcavajani:
@r0ckarong The load balancer has to be setup before deploying the cluster because its IP/FQDN is used during the init of the control plane:

> caaspctl cluster init --control-plane my-loadblancer.example.com milestone3

It has to be configured (before bootstrapping the cluster) to allow traffic for each master that will be used in the cluster.

PR #246, opened 2019-05-31
reboot management and cluster configuration

MaximilianMeister:
this is basically documenting kured and how to configure it in caasp

https://github.com/weaveworks/kured

### DIFF
--- a/adoc/admin-updates.adoc
+++ b/adoc/admin-updates.adoc
@@ -1,13 +1,13 @@
 == Disable automatic updates
 
-Nodes added to a cluster have the service `skuba-update.timer` activated by default. This is the service responsible of running automatic updates. In case you require to completely disable automatic updates for a node it can be done by simply disabling the skuba-update service on the node:
+Nodes added to a cluster have the service `skuba-update.timer`, which is responsible for running automatic updates, activated by default. To disable the automatic updates on a node simply disable the skuba-update service on that node:
 ----
 systemctl disable --now skuba-update.timer
 ----
 
 == Completely disable reboots
 
-If you would like to take care about reboots manually, either temporary or in general, you can disable those by creating a lock:
+If you would like to take care of reboots manually, either as a temporary measure or permanently, you can disable them by creating a lock:
 ----
 kubectl -n kube-system annotate ds kured weave.works/kured-node-lock='{"nodeID":"manual"}'
 ----
--- a/adoc/admin-updates.adoc
+++ b/adoc/admin-updates.adoc
@@ -1,18 +1,18 @@
-=== Disable automatic updates
+== Disable automatic updates
 
 Nodes added to a cluster have the service `skuba-update.timer` activated by default. This is the service responsible of running automatic updates. In case you require to completely disable automatic updates for a node it can be done by simply disabling the skuba-update service on the node:
 ----
 systemctl disable --now skuba-update.timer
 ----
 
-=== Completely disable reboots
+== Completely disable reboots
 
 If you would like to take care about reboots manually, either temporary or in general, you can disable those by creating a lock:
 ----
 kubectl -n kube-system annotate ds kured weave.works/kured-node-lock='{"nodeID":"manual"}'
 ----
 
-=== Manual unlock
+== Manual unlock
 
 In exceptional circumstances, such as a node experiencing a permanent failure whilst rebooting, manual intervention may be required to remove the cluster lock:
 ----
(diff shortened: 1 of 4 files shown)


adoc/deployment-bootstrap.adoc
```diff
@@ -3,13 +3,68 @@
...
 ----
 `cluster init` generates the folder named `my-cluster` and initializes the directory that will hold the configuration for the cluster.
+
+==== Cluster configuration
+
+At this point you can perform some configurations before bootstrapping the cluster.
+
+===== Prevent nodes running special workloads from being rebooted
+
+. Open the `kured` deployment in `my-cluster/addons/kured/kured.yaml`
```
ereslibre: I think that it would be good to explicitly mention that if the cluster is already bootstrapped they can `kubectl edit` the daemonset and achieve the same goal.

MaximilianMeister: true, i almost forgot. thanks for the hint

MaximilianMeister:
added some more hints for prometheus config

adoc/deployment-bootstrap.adoc
```diff
@@ -3,13 +3,92 @@
...
+        ...
+          command:
+            - /usr/bin/kured
+            - --blocking-pod-selector=name=<NAME OF POD>
+----
+
+You can add any key/value labels to this selector:
+----
+--blocking-pod-selector=<LABEL KEY 1>=<LABLE VALUE 1>,<LABEL KEY 2>=<LABEL VALUE 2>
+----
```
lcavajani: @MaximilianMeister This may sounds stupid but I would add the command `kubectl apply -f my-cluster/addons/kured/kured.yaml` so the user knows that it has to be done with `kubectl` and not with `skuba`

adoc/deployment-bootstrap.adoc
```diff
@@ -3,13 +3,92 @@
...
+            - /usr/bin/kured
+            - --prometheus-url=<PROMETHEUS SERVER URL>
+            - --alert-filter-regexp=^(RebootRequired|AnotherBenignAlert|...$
+----
+
+[IMPORTANT]
+====
+The <PROMETHEUS SERVER URL> needs to contain the protocol (`http://` or `https://`)
+====
+
```
lcavajani: same here with `kubectl apply`

adoc/deployment-bootstrap.adoc
MaximilianMeister: this happens before the bootstrap, so `skuba` will do the actual apply during bootstrap. here we just adapt the manifest before skuba reads it

adoc/deployment-bootstrap.adoc
MaximilianMeister: here same thing, skuba will do the apply. the user should never have to `kubectl apply` - only maybe `kubectl edit`

adoc/deployment-bootstrap.adoc
lcavajani: @MaximilianMeister yes but on a running cluster he would have to do it like that. 

What happens if he `kubectl edit` the deployment and then add a new node ?

MaximilianMeister: you can't do `kubectl apply` from the init folder, as there are some variables (e.g. image name) in the yaml that are not expanded (skuba does that during bootstrap)

> What happens if he `kubectl edit` the deployment and then add a new node ?

@ereslibre the join action will apply kured from which manifest?

ereslibre: > @ereslibre the join action will apply kured from which manifest?

Kured is a daemonset and when joining a new node we are not doing anything special with it.

A different story is when upgrading the platform, but that's out of scope right now :)

MaximilianMeister: ok thanks. so kubelet will create the deployment on the new node as it is in the current state in `etcd` - so it should be fine

lcavajani: @MaximilianMeister it works to update the local file and then run `kubectl apply`, I find this is a better way than using `kubectl edit` as the changes are "consistent" between the initial deployed file and the running `ds`

MaximilianMeister: ok makes sense, let's stick to one way of doing it. updated to `kubectl apply`

adoc/admin-updates.adoc
```diff
@@ -0,0 +1,20 @@
+=== Disable automatic updates
+
+Nodes added to a cluster have the service `skuba-update.timer` activated by default. This is the service responsible of running automatic updates. In case you require to completely disable automatic updates for a node it can be done by simply disabling the skuba-update service on the node:
+----
+systemctl disable --now skuba-update.timer
+----
+
```
nkoranova: How do I know which node I'm disabling the update on? Would it maybe make sense to mention that you have to ssh into the node first?

MaximilianMeister: yes indeed. i will add something

MaximilianMeister: @nkoranova updated in the latest commit

PR #268, opened 2019-06-19
Added important note regarding node upgrade before adding to cluster

danielorf:
This ensures that the user only adds nodes to an existing cluster that have matching software versions - this prevents orchestration failure.


nkoranova:
I've put the content in XML in the referenced PR, so I will close this to avoid the confusion. Thanks, @danielorf for your contribution.

PR #273, opened 2019-06-21
Add information about SSH key-based auth

drpaneas:
Currently, the only way to connect via SSH into the openstack
cluster nodes is done via key-based authentication. This is now
mentioned in the docs.

An alternative workaround has been provided changing the
configuration of SSH, enabling password SSH.

### DIFF
--- a/adoc/deployment-ecp.adoc
+++ b/adoc/deployment-ecp.adoc
@@ -22,6 +22,10 @@ source container-openrc.sh
 .. Download the latest SUSE SLES15-SP1 for {soc} from {download_url}.
 .. Upload the image to your {soc}.
 
+.The default user is 'sles'
+[NOTE]
+The SUSE SLES15-SP1 images for {soc} come with pre-defined user `sles` that you use to log in into the cluster nodes. This user has been configured for password-less 'sudo' and is the one recommended to be used by {tf} and `skuba`.
+
 === Deploying the cluster nodes
 
 . Find the {tf} template files for {soc} in `/usr/share/caasp/terraform/openstack`.
@@ -44,11 +48,21 @@ include::deployment-terraform-example.adoc[tags=tf_openstack]
 +
 [TIP]
 ====
-You can set timezone in before deploying the nodes by modifying the files:
+You can set timezone in before deploying the nodes by modifying the file:
 
-* `~/my-deployment/cloud-init/master.tpl`
-* `~/my-deployment/cloud-init/worker.tpl`
+* `~/my-deployment/cloud-init/common.tpl`
 ====
+. Optional: If you absolutely need to be able to ssh into your cluster nodes using password instead of key-based authentication, this is the best time to do it massively for all of your nodes. Alternatively you can do it also later, but manually. To do that you need to modify the cloud-init configuration and comment-out the related SSH configuration:
+* `~/my-deployment/cloud-init/common.tpl`
++
+----
+# Workaround for bsc#1138557 . Disable root and password SSH login
+# - sed -i -e '/^PermitRootLogin/s/^.*$/PermitRootLogin no/' /etc/ssh/sshd_config
+# - sed -i -e '/^#ChallengeResponseAuthentication/s/^.*$/ChallengeResponseAuthentication no/' /etc/ssh/sshd_config
+# - sed -i -e '/^#PasswordAuthentication/s/^.*$/PasswordAuthentication no/' /etc/ssh/sshd_config
+# - systemctl restart sshd
+----
++
 . Enter the registration code for your nodes in `~/my-deployment/registration.auto.tfvars`:
 +
 Substitute `REGISTRATION_CODE` for the code from <<registration_code>>.
@@ -87,4 +101,44 @@ image::deploy-loadbalancer-ip.png[]
 . Now click on menu:Compute[Instances].
 . Switch the filter dropdown to `Instance Name` and enter the string you specified for `stack_name` in the `terraform.tfvars` file.
 . Find the Floating IPs on each of the nodes of your cluster.
+====
+
+=== Log into the cluster notes
+
+. Connecting into the cluster nodes can be accomplished only via SSH key-based authentication thanks to the ssh-public key injection done earlier via {tf}. You can log in as `sles` user.
++
+----
+# With ssh-agent running in the background
+ssh sles@<node-ip-address>
+
+# Without ssh-agent running
+ssh sles@<node-ip-address> -i <path-to-your-ssh-private-key>
+----
+. Once you are connected you can execute commands using password-less `sudo`. In addition to that, you can also set a password if you prefer to.
++
+----
+# Set root password
+sudo passwd
+
+# Set sles user's password
+sudo passwd sles
+----
++
+.Password authentication has been disabled
+[IMPORTANT]
+====
+Even after setting a password for either `root` or `sles` user, you will be unable to log in via SSH using their passwords respectively. Most likely you will receive a `Permission denied (publickey)` error. This mechanism has been disabled on purpose because of security best practices. However, if this environment does not fit your workflows you can change it on your own risk by modifying the SSH configuration on your own.
+
+* `sudo vi /etc/ssh/sshd_config`
+
+
+----
+# Allow password ssh authentication
++ PasswordAuthentication yes
+
+# Allow login as root via ssh
++ PermitRootLogin yes
+----
+
+For the changes to take effect you need to restart the ssh service by running `sudo systemctl restart sshd.service`.
 ====
--- a/adoc/deployment-ecp.adoc
+++ b/adoc/deployment-ecp.adoc
@@ -24,7 +24,7 @@ source container-openrc.sh
 
 .The default user is 'sles'
 [NOTE]
-The SUSE SLES15-SP1 images for {soc} come with pre-defined user `sles` that you use to log in into the cluster nodes. This user has been configured for password-less 'sudo' and is the one recommended to be used by {tf} and `skuba`.
+The SUSE SLES15-SP1 images for {soc} come with pre-defined user `sles`, which you use to log into the cluster nodes. This user has been configured for password-less 'sudo' and is the one recommended to be used by {tf} and `skuba`.
 
 === Deploying the cluster nodes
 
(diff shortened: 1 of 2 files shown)


drpaneas:
This is the missing documentation for https://github.com/SUSE/skuba/pull/387

adoc/deployment-ecp.adoc
```diff
@@ -44,11 +48,21 @@ include::deployment-terraform-example.adoc[tags=tf_openstack]
...
 
-* `~/my-deployment/cloud-init/master.tpl`
-* `~/my-deployment/cloud-init/worker.tpl`
+* `~/my-deployment/cloud-init/common.tpl`
 ====
+. Optional: If you absolutely need to be able to ssh into your cluster nodes using password instead of key-based authentication, this is the best time to do it massively for all of your nodes. Alternatively you can do it also later, but manually. To do that you need to modify the cloud-init configuration and comment-out the related SSH configuration:
+* `~/my-deployment/cloud-init/common.tpl`
++
+----
+# Workaround for bsc#1138557 . Disable root and password SSH login
```
kravciak: I am a bit confused here - why do we hack ssh_config to disable password authentication when sles/root users don't have any password? And then tell customers they can hack it back if they want to set passwords.

I would expect that password authentication is off just by having no password set for system users.
I don't understand what is this feature good for.

Also vmware & bare-metal don't have this config hacks, so why openstack?

drpaneas: First off let me very quickly clarify that is what the team decided when they answered to the question _how would you prefer to be able to ssh to the caasp nodes?_ 

**7 votes** (including the PM): Only with SSH Key based authention. Password-logins should be disabled.
0 votes: Only with password logins.
1 votes: Both.

Now, going to your questions:
> I would expect that password authentication is off just by having no password set for system users.

It is considered best security practice to disable the password authentication for systems running in production. The customer might set a password for either `root` or `sles` but this should not be related to _ssh-ing_ into the cluster remotely. If they want to allow ssh password logins they have to explicitly do it by accepting the security consequences of their actions. If you have password authentication enabled on your SSH service and your root password is weak, chances you are going to get hacked are pretty high. That's why disabling root login is one of the oldest and most used techniques to avoid system compromise on fresh OS installations. You can check this out by leaving the SSH installation by default,parse the `/var/log/secure` file and see how many brute force attacks you will receive within a few hours -- you will be surprised. Even if you are using strong passwords, it's always recommended to disable root login and use a primary SSH user instead, so then if you need to gain root access you can do it using `su` or `sudo` command if required.

It is true that `sles` and `root` users have empty passwords. And this can be a pretty bad thing if you want to keep attackers out of your SSH servers. That's why the best thing you can do is to disable remote logins for accounts with an empty password.

> I don't understand what is this feature good for.

It's hardening the security by mitigating SSH based attacks.

> Also vmware & bare-metal don't have this config hacks, so why openstack?

I guess we have to do it there as well.

adoc/deployment-ecp.adoc
```diff
@@ -88,3 +102,43 @@ image::deploy-loadbalancer-ip.png[]
 . Switch the filter dropdown to `Instance Name` and enter the string you specified for `stack_name` in the `terraform.tfvars` file.
 . Find the Floating IPs on each of the nodes of your cluster.
 ====
+
+=== Log into the cluster notes
+
+. Conneting into the cluster nodes can be accomplished only via SSH key-based authentication thanks to the ssh-public key injection done earlier via {tf}. You can log in as `sles` user.
```
bergmannf: Typo in `conneting` -> `Connecting`

adoc/deployment-ecp.adoc
```diff
@@ -88,3 +102,43 @@ image::deploy-loadbalancer-ip.png[]
...
++
+----
+# With ssh-agent running in the background
+ssh sles@<node-ip-address>
+
+# Without ssh-agent running
+ssh sles@<node-ip-address> -i <path-to-your-ssh-public-key>
```
bergmannf: Slightly incorrect: the `-i` parameter should point to the `private` file: for reference from the man page:

```man
-i identity_file
             Selects a file from which the identity *(private key)* for public key authentication is read.  The default is ~/.ssh/id_dsa, ~/.ssh/id_ecdsa, ~/.ssh/id_ed25519 and
             ~/.ssh/id_rsa.  Identity files may also be specified on a per-host basis in the configuration file.  It is possible to have multiple -i options (and multiple identities
             specified in configuration files).  If no certificates have been explicitly specified by the CertificateFile directive, ssh will also try to load certificate information
             from the filename obtained by appending -cert.pub to identity filenames.
```

adoc/deployment-ecp.adoc
drpaneas: Whoopsie! Thanks for spotting this.

adoc/deployment-ecp.adoc
drpaneas: True. Thanks!

adoc/deployment-ecp.adoc
Martin-Weiss: > 
> 
> First off let me very quickly clarify that is what the team decided when they answered to the question _how would you prefer to be able to ssh to the caasp nodes?_
> 
> **7 votes** (including the PM): Only with SSH Key based authention. Password-logins should be disabled.
> 0 votes: Only with password logins.
> 1 votes: Both.
> 

I have to give a bit feedback as I am not sure how all of this matches to real world ;-).

Did anyone discuss this with customers and what their setup looks like?

So far - in case SSH key based logins are used - the majority of customers to not password protect the private keys that are stored on the servers / jump hosts.
Even up to SES 4 we explained how to set this up and with Salt we also do not have password protected private keys ;-).
AFAIK als SLE-HA also uses non-password protected SSH keys.

> Now, going to your questions:
> 
> > I would expect that password authentication is off just by having no password set for system users.
> 
> It is considered best security practice to disable the password authentication for systems running in production. 
> The customer might set a password for either `root` or `sles` but this should not be related to _ssh-ing_ into the cluster remotely.
> If they want to allow ssh password logins they have to explicitly do it by accepting the security consequences of their actions. 
> If you have password authentication enabled on your SSH service and your root password is weak, chances you are going to get hacked are pretty high. 
> That's why disabling root login is one of the oldest and most used techniques to avoid system compromise on fresh OS installations.
> You can check this out by leaving the SSH installation by default,parse the `/var/log/secure` file and see how many brute force attacks you will receive within a few hours -- you will be surprised. Even if you are using strong passwords, it's always recommended to disable root login and use a primary SSH user instead, so then if you need to gain root access you can do it using `su` or `sudo` command if required.

While all of this is true and should be recommended - the majority of customers do have personalized authentication setup on the servers using i.e. pam with LDAP or AD connection.
Admins login as their user with their LDAP / AD password and then do sudo / ssh to other systems using keys (most often not password protected) or again username + password.


> 
> It is true that `sles` and `root` users have empty passwords. And this can be a pretty bad thing if you want to keep attackers out of your SSH servers. That's why the best thing you can do is to disable remote logins for accounts with an empty password.
> 
> > I don't understand what is this feature good for.
> 
> It's hardening the security by mitigating SSH based attacks.
> 
> > Also vmware & bare-metal don't have this config hacks, so why openstack?
> 
> I guess we have to do it there as well.

IMO we have to support "hardening" and we have to deliver a "hardening guide" similar to what we deliver for SLES - but we also should have defaults and base our solution on what matches to the customers current implementations and security behaviours.

So yes - password protected ssh keys with SSH agent must be supported using non-root users and sudo - but on the other hand also non-password protected and available ssh keys should be used automatically without the need to run ssh-agent and ssh-add for non-password protected private ssh keys.

just my 2 cents ;-)

Martin-Weiss: > 
> 
> First off let me very quickly clarify that is what the team decided when they answered to the question _how would you prefer to be able to ssh to the caasp nodes?_
> 
> **7 votes** (including the PM): Only with SSH Key based authention. Password-logins should be disabled.
> 0 votes: Only with password logins.
> 1 votes: Both.
> 

I have to give a bit feedback as I am not sure how all of this matches to real world ;-).

Did anyone discuss this with customers and what their setup looks like?

So far - in case SSH key based logins are used - the majority of customers to not password protect the private keys that are stored on the servers / jump hosts.
Even up to SES 4 we explained how to set this up and with Salt we also do not have password protected private keys ;-).
AFAIK als SLE-HA also uses non-password protected SSH keys.

> Now, going to your questions:
> 
> > I would expect that password authentication is off just by having no password set for system users.
> 
> It is considered best security practice to disable the password authentication for systems running in production. 
> The customer might set a password for either `root` or `sles` but this should not be related to _ssh-ing_ into the cluster remotely.
> If they want to allow ssh password logins they have to explicitly do it by accepting the security consequences of their actions. 
> If you have password authentication enabled on your SSH service and your root password is weak, chances you are going to get hacked are pretty high. 
> That's why disabling root login is one of the oldest and most used techniques to avoid system compromise on fresh OS installations.
> You can check this out by leaving the SSH installation by default,parse the `/var/log/secure` file and see how many brute force attacks you will receive within a few hours -- you will be surprised. Even if you are using strong passwords, it's always recommended to disable root login and use a primary SSH user instead, so then if you need to gain root access you can do it using `su` or `sudo` command if required.

While all of this is true and should be recommended - the majority of customers do have personalized authentication setup on the servers using i.e. pam with LDAP or AD connection.
Admins login as their user with their LDAP / AD password and then do sudo / ssh to other systems using keys (most often not password protected) or again username + password.


> 
> It is true that `sles` and `root` users have empty passwords. And this can be a pretty bad thing if you want to keep attackers out of your SSH servers. That's why the best thing you can do is to disable remote logins for accounts with an empty password.
> 
> > I don't understand what is this feature good for.
> 
> It's hardening the security by mitigating SSH based attacks.
> 
> > Also vmware & bare-metal don't have this config hacks, so why openstack?
> 
> I guess we have to do it there as well.

IMO we have to support "hardening" and we have to deliver a "hardening guide" similar to what we deliver for SLES - but we also should have defaults and base our solution on what matches to the customers current implementations and security behaviours.

So yes - password protected ssh keys with SSH agent must be supported using non-root users and sudo - but on the other hand also non-password protected and available ssh keys should be used automatically without the need to run ssh-agent and ssh-add for non-password protected private ssh keys.

just my 2 cents ;-)

drpaneas: > So yes - password protected ssh keys with SSH agent must be supported using non-root users and sudo

Great :)


> but on the other hand also non-password protected and available ssh keys should be used automatically without the need to run ssh-agent and ssh-add for non-password protected private ssh keys.

Exactly, you're right. We are still in Beta3 and this feature was not part of the MVP. The idea is that the customer who doesn't want to use `ssh-agent` they should be able to provide the path to their private ssh key as well. While this is easily done in Terraform, we need to change our ssh functionality in skuba. I would like to ask you to open a feature request so we priortize that into the roadmap.

And of course, many thanks for your feedback, it's very welcome to have customer's insights :)

PR #280, opened 2019-06-28
Feature/adoc set test

fsundermeyer:
Created a multipart book from all four CaaS books to support the new "--adocset" feature of DAPS. Works as is with the latest checkout of DAPS (new version has not yet been released).

* On the AsciiDoc side all former books have been squeezed into a sigle book (`set_caasp.adoc`). Each book has become a "part" in the new book.
* Several adjustments (mainly getting the headline levellinmg straight) in the adoc-files were required
* The DC-files have been modified to all use the same MAIN file (`set_caasp.adoc`); individual books are now built via ROOTID
* The DC-file DC-caas-multipart has been added for debugging purposes--it shows the multipart book (without converting it to a DocBook set)

Benefits:

* linking between books is now possible

Caveats:

* asciidoctor is very picky when it comes to headline order. The multipart book requires to strictly follow the respective rules
* reusing files that contain anchors (IDs) in several guides has become more difficult, since anchors (IDs) need to be unique throughout all books (see a half-tested workaround in `deployment_preparation.adoc`)
* when asciidoctor issues non-fatal errors and warnings, the source is usually in a state where the multipart book will not be properly generated. DAPS therefore now refuses to proceed on warnings and errors from asciidoctor


nkoranova:
I have tested this and the feature works well. I haven't found any errors, things look more or less the same as before except that it is now possible to link between the books. I noticed that the Colophon section is now showing up as a separate heading, is that a byproduct of the feature? 
Screenshot from 2019-08-02 15-31-48

r0ckarong:
You need to check out daps from GitHub and run it from there. The following assumes that you check it out to /local/git daps:

1. cd /local/git
2. git clone git@github.com:openSUSE/daps.git
3. Create an alias, e.g. "gdaps": `echo "alias gdaps='/local/git/daps/bin/daps --dapsroot local/git/daps'" >> ~/.alias
4. Source the alias file (will be done automatically whenever opening a new shell, only need to do it now in shells that are currently open): source ~/.alias

r0ckarong:
Added all changes from current master to PR, reworked so it can be merged easily once the DAPS version supporting adocset is released. Setting to blocked for now.

r0ckarong:
Up to now you organized the CaaS documentation into 4 separate books. Each book is a stand-alone book and you cannot link from one book to another. When you build each book, e.g. as PDF, each book has a title page, a table of contents etc.
The change I have done is to combine these 4 separate books into a single book containing 4 parts. You have seen the result when having built it with DC-caasp-multipart.
Theoretically we could stop here and start delivering the complete CaaS doku as such a huge book with parts. If you want to give away the equivalent of nowadays books, you can easily generate it by running for example gdaps -d DC-caasp-multipart pdf --rootid="caas.admin for the Admin Guide. However, the result is not very nice.
The complete book containing everything is too large, and the stand-alone Admin Guide does not have a title page, no table of content and page numbering does not start with 1. Whatsmore it's title is "Part IV: Admin Guide".

What I did is adding an additional conversion step to DAPS. Without that step you will get the result of gdaps -d DC-caasp-multipart pdf (and that is the reason I added this DC-file to the repo - to check what it would like without using the new feature).
With that step (which is acticvated by adding ADOC_SET="yes") to the DC-file, DAPS creates a different XML that contains the four books as real books within a set. A <set> in DocBook is a container into which you can add <book>s and <article>s, so each book/article can "see" the other books/articles and link to them. However, it allows you to build proper books (with title page, table of contents etc.)
What DocBook also allows, is to build parts of documents. You can, for example build a stand-alone chapter of a book. the part you build is identified by it's xml:id which you can pass to DAPS with --rootid. This allows you to build a stand-alone book from a set. If the XML file is called set.xml and the xml:id of the book is "book.one", then you would have a DC-file with MAIN=set.xml and `ROOTID=book.one".
And that is exactly what I have done for CaaS - just look at DC-caasp-admin or the other DC-files. This allows you to create the individual books.

nkoranova:
@fsundermeyer and @r0ckarong this is blocking https://github.com/SUSE/doc-caasp/issues/544 - what can I do to help merge this, so that we can start using links across sets and hence start creating separate guides for each deployment scenario?

r0ckarong:
I think the big reason why I couldn't merge this back when was that the version of DAPS installable from the package channels wasn't able to handle this. Only some dev build of 3.0 could handle it. 

@fsundermeyer Is the version that can handle sets fully available now or please correct me if I misremember this conversation.

r0ckarong:
This is too much efffort to implement with where the product is going.

PR #291, opened 2019-07-01
Skip self-signed certificate verification for VMWare

drpaneas:
SUSE is using _self-signed_ certificate (see ca-certificates-suse)
and this creates problems with Terraform and VSphere provider.

See: https://github.com/hashicorp/terraform/issues/10400

As a result, `terraform plan` is going to crash with the following error:

```
Error: Error refreshing state: 1 error occurred:
	* provider.vsphere: error setting up new vSphere SOAP client: Post https://jazz.qa.prv.suse.net/sdk: x509: certificate signed by unknown authority
```

Most companies provide proper signed certs. However, in case one of our customers do not, then they need to know what the workaround is. I don't know if we need to put that into the docs or not, I have no strong opinion about it.

### DIFF
--- a/adoc/deployment-vmware.adoc
+++ b/adoc/deployment-vmware.adoc
@@ -305,6 +305,10 @@ terraform plan
 terraform apply
 ----
 
+.Skip self-signed certificate verification
+[TIP]
+In case you are using a self-signed certificate you need to skip its verification, as Terraform vsphere provider is not supporting it, thus `terraform plan` will fail. To do this, type the following in your terminal session: `export VSPHERE_ALLOW_UNVERIFIED_SSL=true`
+
 ==== Setup by hand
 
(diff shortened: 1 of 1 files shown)


drpaneas:
The alternative would be to change the configuration for VMWare, so it skips the SSL verification by default, so it goes below the user's radar. What would you prefer?

r0ckarong:
I don't think skipping a default security step is ever a good idea. We are going to document the workaround.

r0ckarong:
This applies to any deployment type and self-signed certs right?

drpaneas:
> This applies to any deployment type and self-signed certs right?

No, this applies only to VMWare deployments afaik, where the Terraform vsphere provider is used. It's a problem of the provider, that doesn't support this functionality.

PR #303, opened 2019-07-03
Remove wrong description for worker

jhsiaosuse:


r0ckarong:
@jhsiaosuse Could you please confirm why this note is no longer true? Will the specified username be created on the system automatically?

jhsiaosuse:
> @jhsiaosuse Could you please confirm why this note is no longer true? Will the specified username be created on the system automatically?

It's wired behind `workers`.

image

jhsiaosuse:
Okay... I think due to OpenStack terraform script removes the username field, so that why I saw it looks strange.

But VMware terraform script still keeps it.

r0ckarong:
> Okay... I think due to OpenStack terraform script removes the username field, so that why I saw it looks strange.
> 
> But VMware terraform script still keeps it.

Aahh, now I understand this. It's a remnant from a removed option. I totally misread the commit. You are right to remove this. My bad. Will merge.

Issue #238, opened 2019-05-28
Create supportconfig docs


r0ckarong:
Relates to: https://github.com/SUSE/avant-garde/issues/287

Should go into "troubleshooting / debugging" in admin guide

jordimassaguerpla:
As a result of the planning, we are moving this to sprint 8.

r0ckarong:
Please ask @flavio and @ereslibre what needs to be done and what info they can provide.

r0ckarong:
Based on the v3 version, describe what `supportconfig` does and what the options are.

https://www.suse.com/documentation/suse-caasp-3/singlehtml/book_caasp_admin/book_caasp_admin.html#cha.admin.troubleshooting

nkoranova:
This will probably require a separate troubleshooting section in the admin guide. We also need to consult the process with CaaSP support.

jordimassaguerpla:
@r0ckarong Can we move this to the Release Coordination project so we have a better overview? Makes sense to you?

jordimassaguerpla:
see https://github.com/SUSE/avant-garde/issues/287#issuecomment-507728278

flavio:
I don't think we have to document anything special, `supportconfig` works in the same way as on SLES. Hence we can point to the official sles documentation, that already has a [section](https://www.suse.com/documentation/sles-12/book_sle_admin/data/sec_admsupport_supportconfig.html) about it (forgive the SLES12 link, that's the first hit I got).

Question for @jordimassaguerpla: are we installing our supportconfig plugin as part of the caasp pattern?

jordimassaguerpla:
Not by default but it is available in the repos.

jordimassaguerpla:
kudos to @mjura , I only accepted it ;)

r0ckarong:
Work is being done in the PR #304, closing.

r0ckarong:
Work is being done in the PR #304, closing.

PR #304, opened 2019-07-03
Add new chapter about troubleshooting

mjura:
Starting new chapter about troubleshooting and the most frequent
problems in CaaSP. Add new section about collecting logs with supportconfig tool.

https://github.com/SUSE/avant-garde/issues/287

### DIFF
--- a/adoc/admin-troubleshooting.adoc
+++ b/adoc/admin-troubleshooting.adoc
@@ -1,0 +1,21 @@
+
+This chapter summarizes frequent problems that can occur while using {productname} and their solutions.
+
+== Overview
+
+This chapter is a collection of frequent problems that are reported for {productname}. Additionally, {suse} support collects problems and their solutions online at link:https://www.suse.com/support/kb/?id=SUSE_CaaS_Platform[] .
+
+In case of problems, a detailed system report can be created with the `supportconfig` command line tool. It will collect information about the system such as: current kernel version, hardware, installed packages, partition setup, and much more. The result is a TAR archive of files. After opening a Service Request (SR), you can upload the TAR archive to Global Technical Support. It will help to locate the issue you reported and to assist you in solving the problem. For details, see link:https://www.suse.com/documentation/sles-12/book_sle_admin/data/cha_adm_support.html[] .
+
+== Collect all relevant logs
+
+To collect all relevant logs run `supportconfig` command on the all master and worker nodes.
+
+[source,bash]
+----
+supportconfig
+tar -xvjf /var/log/nts_*.tbz
+cd nts*
+cat kubernetes.txt crio.txt 
+----
+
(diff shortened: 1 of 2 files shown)


r0ckarong:
@nkoranova This directly relates to https://github.com/SUSE/doc-caasp/issues/238

adoc/admin-troubleshooting.adoc
```diff
@@ -0,0 +1,21 @@
+
+This chapter summarizes frequent problems that can occur while using {productname} and their solutions.
+
+== Overview
+
+This chapter is a collection of frequent problems that are reported for {productname}. Additionally, {suse} support collects problems and their solutions online at link:https://www.suse.com/support/kb/?id=SUSE_CaaS_Platform[] .
+
+In case of problems, a detailed system report can be created with the `supportconfig` command line tool. It will collect information about the system such as: current kernel version, hardware, installed packages, partition setup, and much more. The result is a TAR archive of files. After opening a Service Request (SR), you can upload the TAR archive to Global Technical Support. It will help to locate the issue you reported and to assist you in solving the problem. For details, see link:https://www.suse.com/documentation/sles-12/book_sle_admin/data/cha_adm_support.html[] .
```
nkoranova: Should we add a sentence here that there is no sensitive data among this information? "(...) and much more" sounds a little suspicious.

r0ckarong: Did you confirm there is no sensitive data in what is collected? Since we also collect logs there is likely to be sensitive data in there (user names, error output disclosing sensitive information).

Two good options: 
(Fast) Simply remove the "and so much more" portion since it's colloquial speech anyway and shouldn't be in a document like this.
(Better) Find out what is actually collected and provide a list

I'm fine with either one, if we have the list of stuff collected at hand or can produce it with reasonably low effort then this is preferred.

The phrasing should be changed to something like "and similar information" instead of "much more".

r0ckarong: I've added a link to the support plugin that lists all the data coillected.

adoc/admin-troubleshooting.adoc
```diff
@@ -0,0 +1,46 @@
...
+====
+A full list of of the data collected by `supportconfig` can be found link:https://github.com/SUSE/supportutils-plugin-suse-caasp/blob/master/README.md[here].
+====
+
+To collect all relevant logs run the `supportconfig` command on all the master and worker nodes individually.
+
+[source,bash]
+----
+supportconfig
+tar -xvjf /var/log/nts_*.tbz
```
thehejik: Errata: in my case `supportconfig` produced "txz" archive filetype - so it's not tar.bz2 archive but tar.xz archive. For working with xz archives tar uses `-J` (capital J)

```suggestion
tar -xvJf /var/log/nts_*.txz
```

nkoranova: Thanks, fixed it here: https://github.com/SUSE/doc-caasp/pull/350

PR #343, opened 2019-07-04
Add docs for skuba node upgrade plan

mssola:

### DIFF
--- a/adoc/admin-updates.adoc
+++ b/adoc/admin-updates.adoc
@@ -1,12 +1,19 @@
-== Control Plane
+== Handling updates
 
-In order to get an overview, which cluster-wide updates of the control plane are available, you can run:
+=== Kubernetes components
+
+The update of Kubernetes components is handled via `skuba`.
+
+==== An overview
+
+In order to get an overview of the updates available, you can run:
 
 ----
 skuba cluster upgrade plan
 ----
 
-If the cluster is already running the latest available versions, the output should look like this:
+If the cluster is already running the latest available versions, the output
+should look like this:
 
 ----
 skuba version: 0.6.1 (development) 0a6f78f 20190703 go1.12.4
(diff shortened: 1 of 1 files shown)


mssola:
Blocked by SUSE/skuba#441

r0ckarong:
@mssola Is this unblocked now?

mssola:
@r0ckarong yes

ereslibre:
The new documentation talks about worker nodes, but it's beneath the `Control Plane` section too. I'd revamp this because `skuba cluster upgrade plan` is not strictly `Control Plane` but `Cluster` related, no? Maybe the structure can be fixed in a follow-up PR.

mssola:
@ereslibre @r0ckarong @MaximilianMeister please review again. I've done some restructuring so it's hopefully more clear to the reader which binary does what. Moreover, note that I've added a subsection named "An overview". This is because this section will only tackle the `plan` command. For the `apply` command we should create another subsection.

jordimassaguerpla:
reference https://github.com/SUSE/avant-garde/projects/12

adoc/admin-updates.adoc
```diff
@@ -1,12 +1,19 @@
-== Control Plane
+= Handling updates
```
MaximilianMeister: if you use this `=` header it causes to show an empty page, you have to treat `==` as the largest header here, and use smaller headers from there on. then it renders correctly again

cluster-updates-docs

adoc/admin-updates.adoc
```diff
@@ -1,12 +1,19 @@
...
 
-In order to get an overview, which cluster-wide updates of the control plane are available, you can run:
+== Kubernetes components
```
MaximilianMeister: I would probably also mention the work `Platform` here, so we can easily distinguish between Platform and OS upgrades?

mssola: SUSE CaaS Platform contains both the kubernetes components and the base OS. Here I'm referencing only the components because that's what the skuba commands are covering.

adoc/admin-updates.adoc
mssola: Fixed

PR #346, opened 2019-07-05
Fix RFC5424 message example

innobead:
Add missing double quotes to enclose keys and values of structured data for qualified RFC5424 message format.


innobead:
@r0ckarong or @nkoranova , please help on this PR to align with the fix of https://github.com/SUSE/avant-garde/issues/476. Thanks.

PR #364, opened 2019-07-16
fix quickstart title

eliroca:
### Docserv² Navigation page was not showing the right title for quickstart guide. This should hopefully fix that.


eliroca:
Other maintenance branches may also need this fix.

PR #369, opened 2019-07-18
Clarify sudo requirements on nodes

flavio:
Users must be aware `sudo` must be configured to allow the user specified via `skuba` to work in password-less mode.

This is required to fix [bsc#1140635](https://bugzilla.suse.com/show_bug.cgi?id=1140635).

I also took the chance to fix some minor rendering issues inside of the docs.


flavio:
@nkoranova I've rebased the changes against master and fixed the broken test

PR #381, opened 2019-07-30
[beta] Add terraform SCC and CPI information

mjura:

PR #382, opened 2019-07-30
[master] Add terraform SCC and CPI information

mjura:


mjura:
It was submitted in https://github.com/SUSE/doc-caasp/pull/381

PR #393, opened 2019-08-01
Fix layout of the network ports table

flavio:
The layout of the table got completely messed up when new rows got added by the "row span" rules were not updated to reflect that.

As a result the cells of the table got scrambled, making the table hard to understand.


flavio:
Note well: the table should be simplified and there are information that I honestly question. Let's start by fixing the layout, this will make it easier to review the contents.

I'll follow up with other issues, but let's get this one merged.

PR #398, opened 2019-08-06
Fix network ports description

flavio:
The current contents of the [network ports](https://susedoc.github.io/doc-caasp/master/caasp-quickstart/single-html/#_ports) table are wrong and "bad organized".

The following PR introduces these changes:

  * Ensure we document also the protocol being used by the application
  * Get rid of the requirements that are no longer relevant with v4
  * Remove duplication inside of the table: the table already has a section *"All nodes"*, make consistent usage of it

**TIP for reviewers:** the overall changes are pretty big, I suggest you to review the individual commits to have a better understanding of what has changed. Each commit is well scoped and documented.

**Note well:** I've been able to deploy a cluster on AWS with security rules mapped to the contents of this table. I didn't test dex/gangway, I'll do that later.

## Updated table

I've attached an image of rendered table to make lifer easier for the reviewers.

new ports

## Open questions

I hope I didn't miss anything. Please tell me if some ports should be added/removed.

### CRI-O streaming

I found limited information about it: [here](https://github.com/cri-o/cri-o/blob/master/docs/crio.conf.5.md#crioapi-table)

Is it related with streaming the logs to a central aggregator (CC @innobead and @saschagrunert )?

On a cluster I deployed the `stream_address` is empty, meaning this streaming endpoint is not being used at all.

**Proposal:** let's drop it from the table and document somewhere else its usage (maybe the section about logging). Note well: we should also tell our users the CRI-O stream is **not** going to be secure unless a bunch of `stream_tls_{cert, key, ca}` related options are set.

## Need help reviewing the table

I think the following people can help with the review process:

* @ereslibre 
* @saschagrunert 
* @mrostecki 
* @innobead

### DIFF
--- a/adoc/deployment-sysreqs.adoc
+++ b/adoc/deployment-sysreqs.adoc
@@ -110,7 +110,7 @@ link:https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/hardware
 |===
 |Node |Port |Protocol | Accessibility |Description
 
-.6+|All nodes
+.8+|All nodes
 |22
 |TCP
 |Internal
@@ -146,6 +146,11 @@ link:https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/hardware
 |External
 |Dex (OIDC Connect)
 
+|32001
+|TCP
+|External
+|Gangway (RBAC Authenticate)
+
 .5+|Masters
 |2379
 |TCP
@@ -162,27 +167,6 @@ link:https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/hardware
 |Internal / External
 |Kubernetes API server
 
-|32000
-|TCP
-|External
-|Dex (OIDC Connect)
-
-|32001
-|TCP
-|External
-|Gangway (RBAC Authenticate)
-
-.2+|Workers
-
-|32000
-|TCP
-|External
-|Dex (OIDC Connect)
-
-|32001
-|TCP
-|External
-|Gangway (RBAC Authenticate)
 |===
 
 ==== IP Addresses
--- a/adoc/deployment-sysreqs.adoc
+++ b/adoc/deployment-sysreqs.adoc
@@ -147,7 +147,7 @@ link:https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/hardware
 |Internal
 |VXLAN traffic (used by Cilium)
 
-|10250, 20255
+|10250
 |TCP
 |Internal
 |Kubelet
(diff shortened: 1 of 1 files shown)


saschagrunert:
> ### CRI-O streaming
> I found limited information about it: [here](https://github.com/cri-o/cri-o/blob/master/docs/crio.conf.5.md#crioapi-table)
> 
> Is it related with streaming the logs to a central aggregator (CC @innobead and @saschagrunert )?

Hey, the stream server within CRI-O is used for gRPC streaming endpoints, because [some RPCs of the CRI](https://github.com/kubernetes/cri-api/blob/ba79db1301b40821d3aecf8e92ab4854d7912a00/pkg/apis/runtime/v1alpha2/api.proto#L92-L97) expect such an endpoint to be returned.

If we leave `stream_address = ""`, then the IP of the server will be auto-discovered, so probably the first host IP of the node. If the `stream_address` is commented out, it will default to `127.0.0.1`.

flavio:
@saschagrunert it looks like:

  * We should not ask our users to open this port [EASY]
  * We should also change our default configuration to have `stream_address` either commented or set to `127.0.1`. 

What do you think?

saschagrunert:
> What do you think?

Yes, we could also bind to port `0` to choose any ephemeral one. Let me check if there was any use case where the `stream_address` needs to be non-localhost...

mrostecki:
Apart from @saschagrunert's comment, looks good.

Are there any changes in OpenStack security groups needed?

nirmoy:
cilium-health `4240` port is missing.

mrostecki:
> cilium-health `4240` port is missing.

Yes, I totally forgot about it. It needs to be open for ICMP and TCP connections. On every node.

saschagrunert:
Created a follow-up request in [devel:kubic](https://build.opensuse.org/request/show/721310) to see if my change is acceptable.

flavio:
@mrostecki and @nirmoy according to https://docs.cilium.io/en/v1.4/install/system_requirements/#firewall-rules, `8472` should suffice. What's the scope of port `8471`?

nirmoy:
> @mrostecki and @nirmoy according to https://docs.cilium.io/en/v1.4/install/system_requirements/#firewall-rules, `8472` should suffice. What's the scope of port `8471`?

Isn't `8471` is artifact from flannel's health port ?

flavio:
I've updated the table based on your feedback:

new ports

flavio:
@nkoranova we can merge this change into the docs

PR #405, opened 2019-08-09
Fix cilium-health port description.

nirmoy:
ICMP has no port.


flavio:
@nkoranova you can merge it

PR #407, opened 2019-08-09
Remove comment character in from of the heading

### LINKED ISSUE #371: [doc] 2  Cluster management
This is empty.

2  Cluster management

https://susedoc.github.io/doc-caasp/beta/caasp-admin/single-html/#_cluster_management

bergmannf:
`= Cluster management` heading was starting with a comment.

Fixes #371

### DIFF
--- a/adoc/admin-cluster-management.adoc
+++ b/adoc/admin-cluster-management.adoc
@@ -1,4 +1,4 @@
-//= Cluster Management
+= Cluster Management
 
 Cluster management refers to several processes in the life cycle of a cluster and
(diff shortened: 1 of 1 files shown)


nkoranova:
> `= Cluster management` heading was starting with a comment.
> 
> Fixes #371

That heading was supposed to be commented out (it's a long story, but essentially the only way I found out for the document structure to look ok and build - don't ask). What is actually missing is the other slash to have it properly commented out and it's already in the code, https://github.com/SUSE/doc-caasp/blob/beta/adoc/admin-cluster-management.adoc 
I just have to find out why the build docs haven't changed - that should happen automatically.

Issue #390, opened 2019-07-30
[doc] 3.1.3.1.1  Configuring The Load Balancer


rdannert:
3.1.3.1.1  Configuring The Load Balancer

https://susedoc.github.io/doc-caasp/beta/caasp-deployment/single-html/#_configuring_the_load_balancer

the example nginx.conf used doesn't include configuration options for gangway and dex
as mentioned in the paragraph preceding this one

"It also needs access to Gangway port 32001 and Dex port 32000 on all master and worker nodes in the cluster for RBAC authentication."

these options should be included in the nginx.conf example:

        server {
              listen 32000;
              proxy_pass $upstream;
        }
        server {
              listen 32001;
              proxy_pass $upstream;
        }

jordimassaguerpla:
@kravciak @lcavajani can you confirm we need this configuration?

suseclee:
Working from feature squad: https://github.com/SUSE/avant-garde/issues/762

jordimassaguerpla:
@suseclee Can you provide a PR to doc-caasp?

jordimassaguerpla:
> Working from feature squad: [SUSE/avant-garde#762](https://github.com/SUSE/avant-garde/issues/762)

thanks for picking this one.

PR #418, opened 2019-08-16
Add Dex and gangway ports in cluster load balancer

suseclee:
The example nginx.conf doesn't include configuration options for gangway and dex
as mentioned in the paragraph preceding this one

"It also needs access to Gangway port 32001 and Dex port 32000 on all master and worker nodes in the cluster for RBAC authentication."

Issue: https://github.com/SUSE/doc-caasp/issues/390 

Local nginx LB with this doc was tested and the detailed info was described in https://confluence.suse.com/display/~clee/LB+Skuba+Setup+Doc

### DIFF
--- a/adoc/deployment-loadbalancer.adoc
+++ b/adoc/deployment-loadbalancer.adoc
@@ -101,7 +101,7 @@ stream {
 so the same client will always be redirected to the same server except if this
 server is unavailable.
 <2> Replace the individual `masterXX` with the IP/FQDN of your actual master nodes (one entry each) in the `upstream k8s-masters` section.
-<3> Dex port 32000 and Gangway port 32001 must be accessible through the load balancer for RBAC authentication.
+<3> Dex port 32000 and Gangway port 32001 must be accessible through the load balancer for RBAC authentication
 . Configure `firewalld` to open up port `6443`. As root, run:
 +
 [source,bash]
--- a/adoc/deployment-loadbalancer.adoc
+++ b/adoc/deployment-loadbalancer.adoc
@@ -70,17 +70,46 @@ stream {
         proxy_timeout 3s;
         proxy_pass k8s-masters;
     }
+
+    upstream dex-backends {
+        #hash $remote_addr consistent; // <1>
+        server master00:32000 weight=1 max_fails=1; // <2>
+        server master01:32000 weight=1 max_fails=1;
+        server master02:32000 weight=1 max_fails=1;
+    }
+    server {
+        listen 32000;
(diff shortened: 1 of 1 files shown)


adoc/deployment-loadbalancer.adoc
```diff
@@ -70,17 +70,31 @@ stream {
         proxy_timeout 3s;
         proxy_pass k8s-masters;
     }
+
+    server {
+      listen 32000; // <3>
+      proxy_pass master00:32000; // <2>
```
innobead: Gangway and Dex are exported by NodePort, they are able to access from all nodes. I would suggest to use the same settings as API server as example.

cc @jenting

suseclee: I changed settings for Dex and Gangway. 
@innobead Thanks for the code review.

innobead: This is not I suggested :) 

Like we discussed, you just need to create similar settings of k8s-masters for dex-backends and gangway-backends, then use the correct port (32000, 32001).

jenting: If `master00` down, is dex + gangway still accessible?

suseclee: Somehow, accidentally I rolled back to old version. 
I reverted back to new version.

innobead:
@nkoranova please help merge, thanks!

adoc/deployment-loadbalancer.adoc
```diff
@@ -70,17 +70,49 @@ stream {
...
+        proxy_timeout 3s;
+        proxy_pass gangway-backends; // <3>
+    }
 }
 ----
 <1> **Note:** To enable session persistence, uncomment the `hash` option
 so the same client will always be redirected to the same server except if this
 server is unavailable.
 <2> Replace the individual `masterXX` with the IP/FQDN of your actual master nodes (one entry each) in the `upstream k8s-masters` section.
+<3> Dex port 32000 and Gangway port 32001 need to access througth the load balancer for RBAC authentication
```
nkoranova: I don't quite understand this sentence. :) Perhaps "The Dex port (32000) and the Gangway port (32001) must be accessed through the RBAC load balancer's authentication"?

danielorf: I'd recommend something like:  "Dex port 32000 and Gangway port 32001 must be accessible through the load balancer for RBAC authentication"

The import part being that dex and gangway need to be reachable outside the cluster because the user needs these services to authenticate.

PR #425, opened 2019-08-20
Fix documentation about deletecollection.

### LINKED ISSUE #424: [doc] 3.1.1 deletecollection can remove more than just CronJob
```
deletecollection

    Delete a collection of CronJob.
```

The `deletecollection` verb can be used for more than removing the `CronJob` resource.

bergmannf:
Also explain that it can not be invoked by kubectl but only when using
the k8s API.

Fixes #424

### DIFF
--- a/adoc/admin-security-role-management.adoc
+++ b/adoc/admin-security-role-management.adoc
@@ -22,7 +22,7 @@ delete::
 Delete resources.
 
 deletecollection::
-Delete a collection of CronJob.
+Delete a collection of a resource (can only be invoked using the kubernetes `API`).
 
 get::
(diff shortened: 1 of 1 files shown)


adoc/admin-security-role-management.adoc
```diff
@@ -22,7 +22,7 @@ delete::
 Delete resources.
 
 deletecollection::
-Delete a collection of CronJob.
+Delete a collection of a resource (can only be invoked using the kubernetes `API`).
```
nkoranova:
```suggestion
Delete a collection of a resource (can only be invoked using the {kube} API).
```

PR #430, opened 2019-08-21
Remove obsolete docker and flannel references

lcavajani:
This removes obsolete docker references to kubectl `outputs` and in the `PSP`

Signed-off-by: lcavajani <lcavajani@suse.com>

PR #432, opened 2019-08-21
Fix psp example

bergmannf:
The reference to `docker/default` is replaced with `runtime/default`.


bergmannf:
I just saw already done by #430

PR #436, opened 2019-08-22
Add drain tip for the worker node upgrade (bsc#1145301)

suseclee:
`skuba node upgrade apply` requires restart the services of kubelet and crio.
Then all the pods in the node will be restarted.

`drain` does not have ability to create evicted pods to other nodes. The fact is that drain terminates pods from the node with cordon. But deployment maintains number of pods deployed in a cluster ,so it will create terminated pods to another node due to cordon off the drained node.

This note will show that the customers are aware of restarting the pods. In some special cases, this doc asks the customers to drain the node.


nkoranova:
@ doing the review I felt that there were too many admonitions (note, tip, note, etc.) so I rewrote the section slightly. The information you provided is still there.

PR #443, opened 2019-08-28
Enhance troubleshooting

jenting:

### DIFF
--- a/adoc/admin-troubleshooting.adoc
+++ b/adoc/admin-troubleshooting.adoc
@@ -33,10 +33,10 @@ and worker nodes individually.
 
 [source,bash]
 ----
-supportconfig
-tar -xvJf /var/log/nts_*.txz
-cd nts*
-cat kubernetes.txt crio.txt
+sudo supportconfig
+sudo tar -xvJf /var/log/nts_*.txz
+cd /var/log/nts*
+sudo cat kubernetes.txt crio.txt
 ----
 
 The result is a TAR archive of files. Each of the tar files should be given a name that can be used to identify which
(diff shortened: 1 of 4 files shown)


adoc/admin-troubleshooting.adoc
```diff
@@ -45,4 +45,43 @@ cluster node it was created on.
...
+[source,bash]
+----
+skuba node bootstrap --user sles --sudo --target <IP/FQDN> <NODE NAME> -v=1
+----
+
+== Network Planning And Reconfiguration
+
+It is not possible to reconfigure the used IP ranges of SUSE CaaS Platform once the deployment is complete. Therefore, carefully plan for required IP ranges and future scenarios.
+
+Additionally, keep the network requirements in mind, as stated in Deployment Guide Networking chapter.
```
jenting: @nkoranova I don't know how to link the Deployment Guide Networking chapter.

innobead:
@nkoranova please help review and merge. Also help @jenting question, thanks :+1:

adoc/admin-troubleshooting.adoc
nkoranova: > @nkoranova I don't know how to link the Deployment Guide Networking chapter.

At the moment there is no great way to do that. We can't link between guides. I just write this: `refer to the _{productname} Deployment Guide, Networking chapter_.`.

We have a documentation tool chain feature on the way, that will make this possible though.

nkoranova:
Ok, I've added small formatting edits. I am however not convinced that this is information for the Troubleshooting section. People go to the Troubleshooting section when they are already have the problem and are looking for a way to solve it. Telling them that they should have read the documentation better and that they should try harder next time is IMHO not very helpful and might just make them a bit angry. :D :innocent:  Would you be OK with me adding this as a warning or important notice to the Deployment Guide, Networking chapter instead?

jenting:
> Ok, I've added small formatting edits. I am however not convinced that this is information for the Troubleshooting section. People go to the Troubleshooting section when they are already have the problem and are looking for a way to solve it. Telling them that they should have read the documentation better and that they should try harder next time is IMHO not very helpful and might just make them a bit angry. :D  Would you be OK with me adding this as a warning or important notice to the Deployment Guide, Networking chapter instead?

Agree!

PR #445, opened 2019-08-28
Clarify loadbalancer configuration for upgrades

MaximilianMeister:


adoc/deployment-preparation.adoc
```diff
@@ -121,7 +121,7 @@ Setting up a load balancer is mandatory in any production environment.
 
 {productname} requires a load balancer to distribute workload between the deployed
 master nodes of the cluster. A failure tolerant {productname} cluster will
-always use more than one load balancer since that becomes a "single point of failure".
+always use more than one master node since that becomes a "single point of failure".
```
ereslibre: This was intended, ideally for a real HA environment, aside from several control planes the user should use several load balancers, otherwise if they have a single LB it becomes a single point of failure. Maybe we can rephrase this to state both things: `A failure tolerant {productname} cluster will always use more than one control plane node as well as more than one load balancer, so there is not a single point of failure`, or something similar?

MaximilianMeister: thanks, fixed

nkoranova:
> We use `control plane` and `master`, I think we should tend to use `control plane` in the documentation.

@ereslibre , was this because `master` gets confused with the `master node`, whereas control plane doesn't? I admit to not having a great understanding of this. Reading https://kubernetes.io/docs/concepts/#kubernetes-control-plane I get the impression that Kubernetes Master is a subset of the Kubernetes Control Plane, so they are not completely interchangeable, but i don't fully understand the distinction.

MaximilianMeister:
> was this because `master` gets confused with the `master node`, whereas control plane doesn't? I admit to not having a great understanding of this. Reading https://kubernetes.io/docs/concepts/#kubernetes-control-plane I get the impression that Kubernetes Master is a subset of the Kubernetes Control Plane, so they are not completely interchangeable, but i don't fully understand the distinction.

It's a good point, in this case here if the control plane means the whole set of masters, and a master is only a subset of the control plane, I would stick with `master` nodes as we need to explain things on that level

PR #446, opened 2019-08-28
Missing a reboot command after transactional-update shell

jvanz:
The documentation is missing a reboot command after  files change in a
transactional-update shell. If the customer does not reboot the machine,
the changes will not be used. Possibly, the change will be lost after a
further update due a new snapshot will be created.

Signed-off-by: José Guilherme Vanz <jguilhermevanz@suse.com>


kkaempf:
@nkoranova @jordimassaguerpla - what's the plan on releasing this as a maintenance update ?

PR #450, opened 2019-08-30
Add note about joining nodes sequentially

ereslibre:
We will address this in the future, by allowing worker nodes to be
joined in parallel, and later with the proper tooling, allowing
control plane nodes as well.

### DIFF
--- a/adoc/admin-cluster-management.adoc
+++ b/adoc/admin-cluster-management.adoc
@@ -41,6 +41,11 @@ New master nodes that you didn't initially include in your Terraform's configura
 to be manually added to your load balancer's configuration.
 ====
 
+[IMPORTANT]
+====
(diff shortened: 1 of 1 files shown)


fgerling:
https://bugzilla.suse.com/show_bug.cgi?id=1148988

nkoranova:
Thanks for the PR! It's not great to have two admonitions directly above each other. I would add this bit of info after line 24 and maybe elaborate with a few more words on what "sequential way" means. I understand it as adding them one after another and that you cannot add more of them at the same time.

ereslibre:
> I understand it as adding them one after another and that you cannot add more of them at the same time. If that's what it means, then I'll do the edits. :)

Exactly, that's what it means :) Feel free to perform the edits, thank you!

PR #453, opened 2019-09-02
Added disk provision type

c3y1huang:
Added disk provision type. Default set to `Thick Provision ...`, but requires `Thin Provision`.

Reference: https://github.com/SUSE/avant-garde/issues/807


c3y1huang:
ping/ @nkoranova

PR #464, opened 2019-09-04
Add nfs-client into the SES requirements

### LINKED ISSUE #463: SES requires `ceph-common`, `xfsprogs` and `nfs-client`
see https://bugzilla.suse.com/show_bug.cgi?id=1144991

jordimassaguerpla:
ceph-common and xfsprogs were already in the requirements chapter. We
were missing nfs-client.

This fixes #463 and partially fixes bsc#1144991

Signed-off-by: Jordi Massaguer Pla <jmassaguerpla@suse.de>

### DIFF
--- a/adoc/admin-ses-integration.adoc
+++ b/adoc/admin-ses-integration.adoc
@@ -9,8 +9,8 @@ this chapter describes the steps required for successful integration.
 
 Before you start with integrating {ses}, you need to ensure the following:
 
-* The {productname} cluster must have `ceph-common` and `xfsprogs` installed on all nodes.
-You can check this by running `rpm -q ceph-common` and `rpm -q xfsprogs`.
+* The {productname} cluster must have `ceph-common`, `xfsprogs` and `nfs-client` installed on all nodes.
+You can check this by running `rpm -q ceph-common`, `rpm -q xfsprogs` and `rpm -q nfs-client`.
 * The {productname} cluster can communicate with all of the following {ses} nodes:
 master, monitoring nodes, OSD nodes and the metadata server (in case you need a shared file system).
 For more details refer to the {ses} documentation:


innobead:
This is more about NFS not Ceph support?

jordimassaguerpla:
> This is more about NFS not Ceph support?

https://bugzilla.suse.com/show_bug.cgi?id=1144991 states we need nfs-client

innobead:
> > This is more about NFS not Ceph support?
> 
> https://bugzilla.suse.com/show_bug.cgi?id=1144991 states we need nfs-client

If you check info in the ticket, @lcavajani should mention nfs-client should be about NFS not Ceph. 

```
Like I was saying in GitHub, I understand that we don't want to enforce the usage of Ceph or NFS but I'm 100% we will have complaints/bug reports because ceph.nfs volumes can't be consumed out of the box (like it was in v3). This is already in the documentation but I see this a bad user experience considering many customers have Ceph/NFS.
```

jordimassaguerpla:
> > > This is more about NFS not Ceph support?
> > 
> > 
> > https://bugzilla.suse.com/show_bug.cgi?id=1144991 states we need nfs-client
> 
> If you check info in the ticket, @lcavajani should mention nfs-client is about NFS not Ceph. But we need nfs-client yes for NFS.
> 
> ```
> Like I was saying in GitHub, I understand that we don't want to enforce the usage of Ceph or NFS but I'm 100% we will have complaints/bug reports because ceph.nfs volumes can't be consumed out of the box (like it was in v3). This is already in the documentation but I see this a bad user experience considering many customers have Ceph/NFS.
> ```

Then the docs we have are correct already.

@qnetter I don't see we have any docs about NFS in the admin guide, so I think the best place to mention you need nfs-client installed is in the release notes. Please let us know if you  would like this to be somewhere else other than the release notes.

PR #492, opened 2019-09-18
Remove superfluous productname from guide titles.

nkoranova:
Why is this needed? 
The list of guides at https://documentation.suse.com/suse-caasp/4/ also includes product name and version number for each guide, for example: _SUSE CaaS Platform 4.0.0 Administration Guide_, it should instead only say _Administration Guide_. The `{productname} {productversion}` information is in the subtitle anyway.

### DIFF
--- a/adoc/attributes.adoc
+++ b/adoc/attributes.adoc
@@ -16,7 +16,7 @@
 //Counting upwards from 0, tied to maintenance release
 :productpatch: 0
 :prerelease:
-:productversion: {productmajor}.{productminor}.{productpatch} {prerelease}
+:productversion: {productmajor}.{productminor}.{productpatch}
 :github_url: https://github.com/SUSE/doc-caasp
 
 // Component Versions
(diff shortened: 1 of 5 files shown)


r0ckarong:
Without the additional `{productname/version}` in there the PDF title page will just say _Administration Guide_ without any indication for what product.

This is a workaround because I can't figure out how to force additional content on the title page.

r0ckarong:
This basically is a multi-part issue.

1) Superfluous product name in the listing, that's a docserv2 backend issue, where do they grab the string from that they put in the table
2) We can rework this so the additional names only get inserted when we render PDF
3) Make PDF templating for DAPS easier, currently it involves some convoluted XML template file that takes a while to understand

r0ckarong:
I will merge this for now and see if someone complains about the missing product name on the PDF, to me it's a no-go not having the product on the title page but there is no reasonable technical solution for this right now with DAPS 3.0 and asciidoc.

PR #522, opened 2019-09-24
Add clarification about multiple identities loaded into agent

r0ckarong:


adoc/deployment-preparation.adoc
```diff
@@ -55,11 +55,26 @@ private key for a configurable lifetime. The `-t lifetime` option to
...
 ====
-Skuba will try all the identities loaded into the ssh-agent until one of
-them grants access to the node, or until the ssh server maximum
-authentication attempts are exhausted.
+Skuba will try all the identities loaded into the `ssh-agent` until one of
+them grants access to the node, or until the SSH servers' maximum authentication attempts are exhausted.
+This could lead to undesired messages in SSH or other security/authentication logs on your local machine.
+
+If you wish to avoid authentication attempts with identities irrelevant to {productname}, please configure
+a specific `Host` and `IdentityFile` combination in your local `~/.ssh/config`.
```
dannysauer: Sorry I didn't see the review request earlier. Unfortunately, skuba uses a golang ssh library which does not check the OpenSSH configuration files. So, while we can set OpenSSH to ignore an identity used in skuba, we can not tell skuba to use or ignore any particular identity. This is unfortunate, as it would be nice to be able to specify a user, identity file, etc in the config for skuba. :/

PR #527, opened 2019-09-27
typo: fix kubectl typo

jvanz:
Fixes a typo in the kubectl to get the pods


r0ckarong:
Not a typo. `po` is the shortcut for `pods` in kubectl.

Also this would be a v3 change so the PR must be done against the 'maintenance/CaaS3` branch `master` contains current development (4.x).

Issue #508, opened 2019-09-20
[doc] 5.1.3.1  NGINX Ingress Controller - node port and load balancer instead of external IP


Martin-Weiss:
5.1.3.1  NGINX Ingress Controller

https://documentation.suse.com/suse-caasp/4/single-html/caasp-admin/#_nginx_ingress_controller
----------------------
I believe we should change this:
    externalIPs:
      - 10.86.4.158

--> instead of using external IPs we have to use nodeport and export 30080/30443 and then put a load balancer in front of it (similar to kube-api).

r0ckarong:
@lcavajani Can you please provide the info on how to do this.

Martin-Weiss:
Here is what I have done:

haproxy.conf:
Adjust
```
listen stats
  bind 0.0.0.0:81
  bind :::81 v6only
```

and add the following in /etc/haproxy/haproxy.cfg:

```
frontend nginx-ingress-http
  bind 10.1.1.150:80
  mode tcp
  option tcplog
  default_backend nginx-ingress-http

frontend nginx-ingress-https
  bind 10.1.1.150:443
  mode tcp
  option tcplog
  default_backend nginx-ingress-https

backend nginx-ingress-http
  mode tcp
  option tcp-check
  balance roundrobin
  default-server inter 10s downinter 5s rise 2 fall 2 slowstart 60s maxconn 250 maxqueue 256 weight 100
  server caasp4-05.suse caasp4-05.suse:32080 check
  server caasp4-06.suse caasp4-06.suse:32080 check
  server caasp4-07.suse caasp4-07.suse:32080 check
  server caasp4-08.suse caasp4-08.suse:32080 check

backend nginx-ingress-https
  mode tcp
  option tcp-check
  balance roundrobin
  default-server inter 10s downinter 5s rise 2 fall 2 slowstart 60s maxconn 250 maxqueue 256 weight 100
  server caasp4-05.suse caasp4-05.suse:32443 check
  server caasp4-06.suse caasp4-06.suse:32443 check
  server caasp4-07.suse caasp4-07.suse:32443 check
  server caasp4-08.suse caasp4-08.suse:32443 check

```

And this in values.yaml:

```
diff values.yaml values.yaml.org
259c259
<       #    type: LoadBalancer
---
>     type: LoadBalancer
261c261,266
<     type: NodePort
---
>     # type: NodePort
>     # nodePorts:
>     #   http: 32080
>     #   https: 32443
>     #   tcp:
[...]

PR #530, opened 2019-09-30
Add HAProxy configuration

lcavajani:
I've just added the configuration of HAProxy without the configuration of the HA cluster.

Signed-off-by: lcavajani <lcavajani@suse.com>


r0ckarong:
This should also fix #508

PR #533, opened 2019-10-02
Clusterrolebinding no need to specify namespace

jenting:
Clusterrolebinding no needs to specify the namespace and add the command to on how to list all apiGroups.


nkoranova:
Thanks for the PR.

PR #534, opened 2019-10-03
Fixed cephfs example output

c3y1huang:
Example mounts to rbd, fixed to output mount to monitor nodes.

Reference:
* https://github.com/SUSE/avant-garde/issues/836


c3y1huang:
ping/ @nkoranova , @r0ckarong

nkoranova:
Thnaks!

PR #539, opened 2019-10-04
Extend debugging information

drpaneas:
I extended the documentation related to the missing information we need from the customers when they filling bug reports. This means:

* Highlight the fact they need to provide logs from all the nodes using `IMPORTANT` emphasis
* Add the files for the cluster definition
* Inform them about uploading the `admin.conf` -- _I need your opinion on that_.
* Include the corner case scenario for `skuba remove`
* Inform what to do in case of terraform failure

Github:
- https://github.com/SUSE/avant-garde/issues/897
- https://github.com/SUSE/avant-garde/issues/776

Bugzilla:
- [bsc#1152335](https://bugzilla.suse.com/show_bug.cgi?id=1152335)
- [bsc#1146208](https://bugzilla.suse.com/show_bug.cgi?id=1146208)

PS: We might need to change this after implementing https://github.com/SUSE/avant-garde/issues/566

### DIFF
--- a/adoc/admin-troubleshooting.adoc
+++ b/adoc/admin-troubleshooting.adoc
@@ -57,7 +57,7 @@ debugging purposes. This information is located at the Management node, under yo
 This folder contains important and sensitive information about your {productname} cluster and it's the one
 from where you issue `skuba` commands.
 
-[CAUTION]
+[WARNING]
 ====
 If the problem you are facing is related to your production environment, do **not** upload the `admin.conf` as this would
 expose access to your cluster to anyone in possession of the collected information!
--- a/adoc/admin-troubleshooting.adoc
+++ b/adoc/admin-troubleshooting.adoc
@@ -34,7 +34,6 @@ To collect all the relevant logs, run the `supportconfig` command on all the mas
 and worker nodes individually.
 ====
 
-
 [source,bash]
 ----
 sudo supportconfig
@@ -46,7 +45,7 @@ sudo cat kubernetes.txt crio.txt
 The result is a `TAR` archive of files. Each of the `*.txz` files should be given a name that can be used to identify which
 cluster node it was created on.
 
-After opening a Service Request (SR), you can upload the TAR archives to Global Technical Support.
+After opening a Service Request (SR), you can upload the `TAR` archives to {suse} Global Technical Support.
 
 The data will help to debug the issue you reported and assist you in solving the problem.
 For details, see https://www.suse.com/documentation/sles-15/book_sle_admin/data/cha_adm_support.html.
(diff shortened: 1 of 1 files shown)


adoc/admin-troubleshooting.adoc
```diff
@@ -39,15 +43,49 @@ cd /var/log/nts*
...
+== Cluster definition directory
+
+Apart from the logs provided by running the `supportconfig` tool, an additional set of data might be required for
+debugging purposes. This information is located at the Management node, under your cluster definition directory.
+This folder contains important and sensitive information about your CaaSP cluster and it's the one
+from where you issue `skuba` commands.
+
+[TIP]
+====
+If the problem you are facing is related to your production environment, do **not** upload the `admin.conf` as this would
```
ereslibre: We should also advise to not upload the `pki` directory, since this also contains sensitive information (ca cert and key).

adoc/admin-troubleshooting.adoc
```diff
@@ -57,6 +95,22 @@ For example, if bootstraps the first master node of the cluster fails, re-run th
 skuba node bootstrap --user sles --sudo --target <IP/FQDN> <NODE NAME> -v=1
 ----
 
+However if the bootstrap or join procedure fails at the last final steps, re-running it might not help, giving you an error similar to this:
+
+[source,bash]
+----
+error execution phase control-plane-join/update-status: error uploading configuration: unable to create ConfigMap: etcdserver: request timed out
```
ereslibre: What if instead of referring to this esoteric message we describe that removing the node needs to be done if `kubectl get nodes` shows the node on the list.

adoc/admin-troubleshooting.adoc
drpaneas: I forgot about it. Thanks for the sharp eye ;)

adoc/admin-troubleshooting.adoc
```diff
@@ -57,6 +98,25 @@ For example, if bootstraps the first master node of the cluster fails, re-run th
 skuba node bootstrap --user sles --sudo --target <IP/FQDN> <NODE NAME> -v=1
 ----
 
+However, if the `bootstrap` (or `join`) procedure fails at the last final steps, re-running it might _not_ help. To verify
```
ereslibre: I think we can reduce this case to `join` only. Bootstrap is safe to be retried, as we'll just `kubeadm reset`.

Also, regarding join, it would be good to know under what conditions this can fail and if we can do something about it (if skuba detects that the join was partially made, as part of the "rollback" operation it could delete the node from the cluster if it happened to register, and possibly perform other tasks).

drpaneas: To be on the safe side, I would propose to leave it as it is for now. Running `kubeadm` commands outside of `skuba` sounds scary to me, so I would avoid telling the customers to interact with kubeadm but use only skuba as their interface.

But, you are right. This is not a proper fix. Thus, I would propose I could create another GH where we could focus on a better solution, addressing the root problem and involve QA as well.

ereslibre: Oh no no, I mean that we are [literally doing a `kubeadm reset` with skuba](https://github.com/SUSE/skuba/blob/f8534b1bc369335870f948715ee7f0a446df0b2d/pkg/skuba/actions/node/bootstrap/bootstrap.go#L137), so the operation can be retried.

drpaneas: I see, yes that makes sense. I narrowed down the scope only to `join`.

Issue #537, opened 2019-10-04
[doc] Wrong certificate path on CPI instructions


kravciak:
If certificate is uploaded to /etc/ssl/certs it disappears after cluster update. This means kubernetes looses access to openstack and cluster dies.

**What we have:**
[3.2.1  Deploying the Cluster Nodes](https://documentation.suse.com/suse-caasp/4/single-html/caasp-deployment/#_deploying_the_cluster_nodes)
```
# Used to specify the name of to your custom CA file located in /etc/ssl/certs.
# Upload CUSTOM_CA_FILE to this path on nodes before joining them to your cluster.
#ca_file = "/etc/ssl/certs/<CUSTOM_CA_FILE>"
```
[4.2.3.2  Example OpenStack Cloud Provider Configuration](https://documentation.suse.com/suse-caasp/4/single-html/caasp-deployment/#_example_openstack_cloud_provider_configuration)
```
ca-file="/etc/ssl/certs/SUSE_Trust_Root.pem"
```

**What we should have:**
[3.2.1  Deploying the Cluster Nodes](https://documentation.suse.com/suse-caasp/4/single-html/caasp-deployment/#_deploying_the_cluster_nodes)
```
# Used to specify the name of to your custom CA file located in /etc/pki/trust/anchors/.
# Upload CUSTOM_CA_FILE to this path on nodes before joining them to your cluster.
# ca_file = "/etc/pki/trust/anchors/<CUSTOM_CA_FILE>"
```
[4.2.3.2  Example OpenStack Cloud Provider Configuration](https://documentation.suse.com/suse-caasp/4/single-html/caasp-deployment/#_example_openstack_cloud_provider_configuration)
```
ca-file="/etc/pki/trust/anchors/SUSE_Trust_Root.pem"
```

Martin-Weiss:
Should this: ca-file="/etc/pki/trust/anchors/SUSE_Trust_Root.pem" not be /etc/ssl/ca-bundle.pem ?
-> after we copy the public cert of the CA to /etc/pki/trust/anchors/SUSE_Trust_Root.pem and execute "update-ca-certificates" it is going to /etc/ssl/ca-bundle.pem  which includes all CAs that are trusted on the system.

kravciak:
ca-file is [used here](https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#global) as kubernetes config, not sure if we want ca-bundle.pem

But I am no expert here - @mjura maybe knows

[...]

PR #540, opened 2019-10-04
Fix certificate path in CPI instructions

kravciak:
If certificate is uploaded to /etc/ssl/certs it disappears after cluster update. This means kubernetes looses access to openstack and cluster dies.

PR for https://github.com/SUSE/doc-caasp/issues/537


nkoranova:
thanks!

PR #547, opened 2019-10-08
Add info about disabling IPv6 bsc#1152810

lcavajani:
Here are the info on how to disable IPv6 without encouraging it...

@nkoranova @r0ckarong there is an issue with `cilium.disable_ipv6` reference, do you have an idea how to fix this ?

```
Building DC-caasp-admin
daps -d /daps_temp-98g5jg6V02/source/DC-caasp-admin html  
daps -d /daps_temp-98g5jg6V02/source/DC-caasp-admin pdf  
Building DC-caasp-architecture
daps -d /daps_temp-98g5jg6V02/source/DC-caasp-architecture html  
daps -d /daps_temp-98g5jg6V02/source/DC-caasp-architecture pdf  
Building DC-caasp-deployment
daps -d /daps_temp-98g5jg6V02/source/DC-caasp-deployment html  
daps -d /daps_temp-98g5jg6V02/source/DC-caasp-deployment pdf  
Building DC-caasp-quickstart
/daps_temp-98g5jg6V02/source/build/.profiled/noprofile/caasp-quickstart.xml:367:89: error: IDREF "cilium.disable_ipv6" without matching ID
make: *** [validate] Error 1
/usr/share/daps/make/validate.mk:40: recipe for target 'validate' failed
Error: exit status 2
(Exiting d2d_runner) DC-caasp-quickstart has validation issues and cannot be built.
(Exiting d2d) Oh no. There are no documents for you.
```

https://github.com/SUSE/avant-garde/issues/920

Fixes bsc#1152810

Signed-off-by: lcavajani <lcavajani@suse.com>

### DIFF
--- a/adoc/deployment-cilium.adoc
+++ b/adoc/deployment-cilium.adoc
@@ -41,7 +41,7 @@ https://cilium.readthedocs.io/en/v1.5/gettingstarted/memcached/
 Making security policies on clusters on AWS::
 https://cilium.readthedocs.io/en/v1.5/gettingstarted/aws/
 
-[[cilium.disable_ipv6]]
+[[cilium-disable-ipv6]]
 ==== Disable IPv6
 
 . Switch to the cluster file directory (e.g `my-cluster`).
--- a/adoc/deployment-cilium.adoc
+++ b/adoc/deployment-cilium.adoc
@@ -40,3 +40,40 @@ https://cilium.readthedocs.io/en/v1.5/gettingstarted/memcached/
 
 Making security policies on clusters on AWS::
 https://cilium.readthedocs.io/en/v1.5/gettingstarted/aws/
+
+[[cilium.disable_ipv6]]
+==== Disable IPv6
+
+. Switch to the cluster file directory (e.g `my-cluster`).
+
+. In the cilium file `./addons/cni/cilium.yaml`, edit the args for the container
+with the name `cilium-agent` within the DaemonSet `cilium` to add "--enable-ipv6=false".
+This part should look like this (the args are subject to changes):
++
+[source,yaml]
+----
+      containers:
+      - image: registry.suse.com/caasp/v4/cilium
+        imagePullPolicy: IfNotPresent
+        name: cilium-agent
+        command: [ "cilium-agent" ]
+        args:
+          - "--debug=$(CILIUM_DEBUG)"
+          - "--disable-envoy-version-check"
+          - "-t=vxlan"
+          - "--kvstore=etcd"
+          - "--kvstore-opt=etcd.config=/var/lib/etcd-config/etcd.config"
+          - "--disable-ipv4=$(DISABLE_IPV4)"
+          - "--container-runtime=crio"
+          - "--enable-ipv6=false"
+----
+
+. There is nothing to do if this configuration is modified before the
+bootstrap of the cluster as it will be automatically applied.
+If the cluster is already bootstrapped, it is required to apply the changes:
++
+[source,bash]
+----
+kubectl apply -f ./addons/cni/cilium.yaml
+----
+The cilium pods will then be recreated one by one with the new parameters.
--- a/adoc/deployment-cilium.adoc
+++ b/adoc/deployment-cilium.adoc
@@ -1,3 +1,4 @@
+
 == Network Security
 
 === Cilium
--- a/adoc/deployment-sysreqs.adoc
+++ b/adoc/deployment-sysreqs.adoc
@@ -229,10 +229,12 @@ This is not sufficient because Cilium will still try to enable IPv6 since it is 
 The first option will result in IPv6 not being disabled and the second will lead to Cilium not being able to start.
 ====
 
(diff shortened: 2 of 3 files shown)


lcavajani:
cc @ereslibre

innobead:
@ereslibre you mean when executing addon update it will cause customized manifests overrided by the current implementation?

lcavajani:
@innobead The customization of the manifests is part of the whole addon upgrade but it is still not created. So maybe it's time to prioritized :)

(I assume in this doc that until we ship `1.16` the addon upgrade with kustomize would be ready)

innobead:
> @innobead The customization of the manifests is part of the whole addon upgrade but it is still not created. So maybe it's time to prioritized :)
> 
> (I assume in this doc that until we ship `1.16` the addon upgrade with kustomize would be ready)

@lcavajani thanks for the info and yes, it looks high priority to me.

@jordimassaguerpla as u have an option to plan to release from master, please follow up this (customized manifest keep) to avoid overriding the customers deployment by 1.0.2 until it gets implemented. Thanks.

adoc/deployment-sysreqs.adoc
```diff
@@ -203,13 +203,39 @@ The management workstation needs at least the following networking permissions:
...
 
-==== IP Addresses
+==== IPv6
+
+[WARNING]
+====
+Using IPv6 addresses is currently not supported but it is not disabled.
+====
+
+A common practice to disable IPv6 at the node level is to configure the following kernel parameter:
```
kkaempf: Where does this need to be configured ? Mention the exact file that needs to be adapted.

lcavajani: I didn't want to write a guide on how to disable IPv6 so I just refer to common practices used by sysadmins. IMO we should not encourage disabling IPv6.

lcavajani: I mean disabling it at the host level*

adoc/deployment-cilium.adoc
```diff
@@ -40,3 +40,40 @@ https://cilium.readthedocs.io/en/v1.5/gettingstarted/memcached/
...
 Making security policies on clusters on AWS::
 https://cilium.readthedocs.io/en/v1.5/gettingstarted/aws/
+
+[[cilium-disable-ipv6]]
+==== Disable IPv6
+
+. Switch to the cluster file directory (e.g `my-cluster`).
+
+. In the cilium file `./addons/cni/cilium.yaml`, edit the args for the container
+with the name `cilium-agent` within the DaemonSet `cilium` to add "--enable-ipv6=false".
```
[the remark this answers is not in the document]
lcavajani: We could just disable it for every cluster but not sure this is the right path IMHO.

adoc/deployment-sysreqs.adoc
kkaempf: > IMO we should not encourage disabling IPv6.

Well, customers do disable IPv6 for "security reasons" ;-) and don't care about our recommendations :-/

ereslibre:
> @ereslibre you mean when executing addon update it will cause customized manifests overrided by the current implementation?

Yes, exactly. If we evolve the addon configuration/version bump, we will blindly reapply the new configuration, overriding the customer customizations.

Using `kustomize` would allow us to evolve our base while keeping customer customizations on top of it. I will open a RFC really soon tackling this.

adoc/deployment-sysreqs.adoc
Martin-Weiss: It is a best practice and recommended to turn of all protocols that are not required and not used.
(not only for our systems ;-)) 
That is the reason why at least consulting recommends to turn of IPv6. 

As we do not support IPv6 with CaaSP / K8S at the moment and AFAIK K8S does not support dual-stack - I believe the default should be not to "depend" on it. 

Once we have IPv6 support this should be changed or at least the decision needs to be done during the deployment with Skuba.

lcavajani: Dual stack is in alpha in `1.16`.

We can just turn this into "enabling IPv6" if we go with disabling by default in skuba. No strong opinions on this.

jordimassaguerpla:
> > @innobead The customization of the manifests is part of the whole addon upgrade but it is still not created. So maybe it's time to prioritized :)
> > (I assume in this doc that until we ship `1.16` the addon upgrade with kustomize would be ready)
> 
> @lcavajani thanks for the info and yes, it looks high priority to me.
> 
> @jordimassaguerpla as u have an option to plan to release from master, please follow up this (customized manifest keep) to avoid overriding the customers deployment by 1.0.2 until it gets implemented. Thanks.

Sorry I don't understand. Can you elaborate a bit more?

adoc/deployment-sysreqs.adoc
Martin-Weiss: Can´t we enhance the cilium pod to autodetect if ipv6 is available?
(sounds a bit like working around a missing feature in that pod ;-))

lcavajani:
Closing as `IPv6` is now disabled by default.

PR #576, opened 2019-10-29
Fix etcdctl cluster-health command

paraenggu:
The following PR fixes the `etcdctl cluster-health` instructions, as this command doesn't work with `ETCDCTL_API=3`:

```
Error: unknown command "cluster-health" for "etcdctl"
```

Simply removing the export will result in using the v2 API, as this will be defined while sourcing the `etcdctl` environment file.

Furthermore, the PR only turns the auto export attribute (`set -a`) on during sourcing to prevent subsequent variables from being exported automatically.

### DIFF
--- a/xml/admin_troubleshooting.xml
+++ b/xml/admin_troubleshooting.xml
@@ -267,7 +267,6 @@ entrypoint.sh bundle exec rails runner \
   </para>
 <screen>&prompt.root;<command>set -a</command>
 &prompt.root;<command>source /etc/sysconfig/etcdctl</command>
-&prompt.root;<command>export ETCDCTL_API=3</command>
 &prompt.root;<command>etcdctl cluster-health</command></screen>
  </sect1>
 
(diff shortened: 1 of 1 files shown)


r0ckarong:
@ereslibre @lcavajani Was there a specific reason we forced API v3 here?

lcavajani:
Yes, let's stick with the new version of the API. In `v3`, cluster-health does not exist anymore but it has been replaced. You now have to use: `etcdctl endpoint status --cluster`.

r0ckarong:
@paraenggu Please make sure to enable maintainer updates on future pull requests. I can't push corrective commits into your branch. I will do the change separately this time and close this issue.

r0ckarong:
Closed in https://github.com/SUSE/doc-caasp/commit/8a3c02126d2a2a10cf7d3ca7f243913a683f151a

lcavajani:
@r0ckarong in case you need an output, it looks like this:

```
# etcdctl --cacert=/etc/kubernetes/pki/etcd/ca.crt --cert=/etc/kubernetes/pki/etcd/peer.crt --key=/etc/kubernetes/pki/etcd/peer.key endpoint status --cluster
https://192.168.0.1:2379, 34545803791e2f10, 3.3.11, 5.6 MB, true, 105, 10872508
https://192.168.0.2:2379, b5641ccfa1ba794e, 3.3.11, 5.8 MB, false, 105, 10872508
https://192.168.0.3:2379, f4ceeaa452ae1ad2, 3.3.11, 5.6 MB, false, 105, 10872508

PR #583, opened 2019-11-04
Add missing steps for machine ID generation on manual VMWare deployment (bsc#1155322)

r0ckarong:


adoc/deployment-vmware.adoc
```diff
@@ -275,19 +275,32 @@ terraform plan
...
 
+==== Setup by Hand
+Deploy the template to your created VMs. After that, boot into the node and configure
+the OS as needed.
+
+.Regenerating Machine ID
 [IMPORTANT]
 ====
-Make sure to give each VM in VMware a clear name that shows its purpose in the cluster, for example
+During the template preparation you will have removed the machine ID from the template image.
```
thehejik: Please move this para about regeneration of machine-id into "Deploying VMs from the Template" section and introduce the para by something like:
```In case you are using template but NOT terraform for deployment please do:```

File: adoc/deployment-vmware.adoc [hunk 2]
```diff
@@ -275,19 +275,32 @@ terraform plan
 terraform apply
 ----
 
-==== Setup by Hand
+==== Setup with {ay}
```
thehejik: This heading should be one level up (it is not related to templates) : `=== Setup with {ay}`

PR #589, opened 2019-11-11
Remove superfluous Handling Updates heading.

nkoranova:
I am not sure what the purpose of that section was, removing it flattens the structure and makes it more transparent.

### DIFF
--- a/adoc/admin-updates.adoc
+++ b/adoc/admin-updates.adoc
@@ -1,11 +1,10 @@
 [[handling_updates]]
-== Handling Updates
 
-=== Updating Kubernetes Components
+== Updating Kubernetes Components
 
 Updating of {kube} components is handled via `skuba`.
 
-==== Generating an Overview of Available Platform Updates
+=== Generating an Overview of Available Platform Updates
 
 In order to get an overview of the updates available, you can run:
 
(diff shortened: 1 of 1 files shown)


r0ckarong:
There are three distinct subsections:
* Updating Kubernetes Components
* (Single) Node Updates
* SLES Updates (Base OS)

those are three related (but different) topics that all need to be describe under "Handling Updates". I would assume this was intentional when Max wrote it.

This PR would remove the hierarchy and flatten it all under "Updating Kubernetes Components" which is wrong.

nkoranova:
> There are three distinct subsections:
> 
>     * Updating Kubernetes Components
> 
>     * (Single) Node Updates
> 
>     * SLES Updates (Base OS)
> 
> those are three related (but different) topics that all need to be described under "Handling Updates". I would assume this was intentional when Max wrote it.

I wrote this section with Max and I checked with him on this.

> This PR would remove the hierarchy and flatten it all under "Updating Kubernetes Components" which is wrong.

That's not true.  The PR only removes the Handling Updates heading, and it creates three distinct subsections directly under the Cluster Updates chapter. 

image

r0ckarong:
Sorry, I misread this. My bad.

PR #591, opened 2019-11-12
Fix Dex configuration is broken after update (#1014)

cclhsu:
# Describe your changes
1. Instead of modify Dex connector using addon after deploy, edit configmap directly.

### DIFF
--- a/adoc/admin-security-ldap.adoc
+++ b/adoc/admin-security-ldap.adoc
@@ -455,7 +455,7 @@ Example LDIF configuration to create group `k8s-admins` using an LDAP command:
   objectCategory: cn=Group,cn=Schema,cn=Configuration,dc=example,dc=org
 ====
 
-==== Example 2: Dex Active Directory TLS Connector Configuration (addons/dex/dex.yaml)
+==== Example 2: Use command `kubectl --namespace=kube-system edit configmap oidc-dex-config` to edit Dex ConfigMap.
 Dex connector template configured to use Active Directory:
 ----
 connectors:
--- a/adoc/admin-security-ldap.adoc
+++ b/adoc/admin-security-ldap.adoc
@@ -455,7 +455,8 @@ Example LDIF configuration to create group `k8s-admins` using an LDAP command:
   objectCategory: cn=Group,cn=Schema,cn=Configuration,dc=example,dc=org
 ====
 
-==== Example 2: Use command `kubectl --namespace=kube-system edit configmap oidc-dex-config` to edit Dex ConfigMap.
+==== Example 2: Dex Active Directory TLS Connector Configuration
+Run `kubectl --namespace=kube-system edit configmap oidc-dex-config` to edit Dex ConfigMap.
 Dex connector template configured to use Active Directory:
 ----
 connectors:
--- a/adoc/admin-security-ldap.adoc
+++ b/adoc/admin-security-ldap.adoc
@@ -457,7 +457,7 @@ Example LDIF configuration to create group `k8s-admins` using an LDAP command:
 
 ==== Example 2: Dex Active Directory TLS Connector Configuration
 Run `kubectl --namespace=kube-system edit configmap oidc-dex-config` to edit Dex ConfigMap.
-Dex connector template configured to use Active Directory:
+Configure Dex ConfigMap to use Active Directory using the following template:
 ----
 connectors:
 - type: ldap
--- a/adoc/admin-security-rbac.adoc
+++ b/adoc/admin-security-rbac.adoc
@@ -92,7 +92,7 @@ kubectl create rolebinding admin --clusterrole=admin --user=<USER-1> --user=<USE
 Administrators can update the authentication connector settings after {productname}
 deployment as follows:
 
-. Run the following `kubectl` command to access to access Dex ConfigMap:
+. Run the following `kubectl` command to access Dex ConfigMap:
 +
 ----
 kubectl --namespace=kube-system edit configmap oidc-dex-config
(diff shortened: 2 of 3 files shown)


adoc/admin-security-ldap.adoc
```diff
@@ -455,7 +455,8 @@ Example LDIF configuration to create group `k8s-admins` using an LDAP command:
   objectCategory: cn=Group,cn=Schema,cn=Configuration,dc=example,dc=org
 ====
 
-==== Example 2: Dex Active Directory TLS Connector Configuration (addons/dex/dex.yaml)
+==== Example 2: Dex Active Directory TLS Connector Configuration
```
nkoranova: @cclhsu I've changed the Example 2 heading you had there because I didn't understand the content and how it relates to the example and what the steps are the user should take. Is the text below 'Dex connector template configured to use Active Directory:' what the user should add to the Dex ConfigMap?

cclhsu: Dex connector can be configured to use different authentication service.  The example template is from Dex which provides user with detail descriptions on how to configure the Dex configmap using Active Directory.

adoc/admin-security-rbac.adoc
```diff
@@ -134,10 +139,13 @@ Besides the LDAP connector you can also set up other connectors.
 For additional connectors, refer to the available connector configurations in the Dex repository:
 https://github.com/dexidp/dex/tree/v2.16.0/Documentation/connectors.
 +
-. Update Dex deployment:
+. Save and exit Dex ConfigMap by typing `:wq` in the terminal.
```
r0ckarong: This is only needed when using `vi/vim` to edit the file. We don't specify the use of this editor. Please rephrase to "Save the changed ConfigMap".

adoc/admin-security-rbac.adoc
```diff
@@ -92,8 +92,13 @@ kubectl create rolebinding admin --clusterrole=admin --user=<USER-1> --user=<USE
 Administrators can update the authentication connector settings after {productname}
 deployment as follows:
 
-. Open the `dex` configmap in `<cluster name>/addons/dex/dex.yaml` .
-. Adapt ConfigMap by adding LDAP configuration to the connector section of the `config.yaml` data.
+. Run the following `kubectl` command to access to access Dex ConfigMap:
```
r0ckarong: Double "to access". 

Should be reworded to "... command to access the Dex ConfigMap:"

adoc/admin-security-rbac.adoc
thehejik: I would rather let it be as it is because `kubectl edit configmap whatever` is opening the file in `vim` by default (or editor set in $EDITOR env. variable but by default it's `vim`). 

With the proposed change some non-unix users might be confused and will not know how to save&exit.

adoc/admin-security-rbac.adoc
```diff
@@ -134,10 +139,13 @@ Besides the LDAP connector you can also set up other connectors.
...
 +
-. Update Dex deployment:
+. Save and exit Dex ConfigMap by typing `:wq` in the terminal.
++
+. Restart Dex and Gangway by running:
 +
 ----
-kubectl replace --force -f <cluster-name>/addons/dex/dex.yaml
+kubectl --namespace=kube-system delete pod -l app=oidc-dex
+kubectl --namespace=kube-system delete pod -l app=oidc-gangway
```
jenting: I don't think we need to restart gangway pods since this is for fix dex configuration.

jenting:
@cclhsu Sorry I don't know the discussion history, but is it better to ask the customer to download the existing `oidc-dex-config` ConfigMap before addon upgrade, then re-apply the downloaded `oidc-dex-config` ConfigMap after addon upgraded. Since as I know, we did not change the `oidc-dex-config` ConfigMap since skuba `4.0.0`.

```
kubectl get configmap oidc-dex-config -n kube-system -o yaml > oidc-dex-config.yaml
kubectl replace -f oidc-dex-config.yaml
```

adoc/admin-security-ldap.adoc
nkoranova: Ok, thanks, I've edited the bit below to make that a little clearer.

adoc/admin-security-rbac.adoc
nkoranova: I agree with @thehejik and would leave that part in. How likely is it that people will go change the $EDITOR env. variable on the virtual machine?

r0ckarong: OK agreed. If this opens vim automatically then this makes sense.

nkoranova:
@jordimassaguerpla is this relevant for 4.0.2?

davidcassany:
> @jordimassaguerpla is this relevant for 4.0.2?

Yes it is, this makes reference to SUSE/avant-garde#1014

nkoranova:
> @cclhsu Sorry I don't know the discussion history, but is it better to ask the customer to download the existing `oidc-dex-config` ConfigMap before addon upgrade, then re-apply the downloaded `oidc-dex-config` ConfigMap after addon upgraded. Since as I know, we did not change the `oidc-dex-config` ConfigMap since skuba `4.0.0`.
> 
> ```
> kubectl get configmap oidc-dex-config -n kube-system -o yaml > oidc-dex-config.yaml
> kubectl replace -f oidc-dex-config.yaml
> ```

I asked Clark about this in a private message and apparently both ways are OK, it seems to be a matter of preference. So closing and merging now. Thanks for the contribution!

r0ckarong:
@nkoranova Can you merge this to master as well please?

PR #599, opened 2019-11-15
Add information about subnet sizing (Closes #593)

### LINKED ISSUE #593: [doc] 5.2.2.1  Network Settings -> add a recommendation for size of service and pod network
5.2.2.1  Network Settings

https://susedoc.github.io/doc-caasp/master/caasp-deployment/single-html/#_network_settings

---
I would recommend to specify /16 or bigger for the pod / overlay network (/16 allows up to 256*256 pods)
And I would recommend to specify a smaller network for the services as in general the number of services can be expected to be smaller than the number of pods - but probably never bigger.
So for services something between /16 and /24 (but not /12 which would be more services than pods ;-)).

r0ckarong:
Closes #593 
Add information about the subnet sizing.

### DIFF
--- a/adoc/deployment-sysreqs.adoc
+++ b/adoc/deployment-sysreqs.adoc
@@ -168,7 +168,7 @@ Please plan generously for workload and the expected size of the networks before
 The default pod subnet is `10.244.0.0/16`. It allows for 65536 IP addresses overall.
 Assignment of CIDR's is by default `/24` (254 usable IP addresses per node).
 
-The default node allocation of `/24` means a hard cluster node limit of 254 since this is the number of `/24` ranges that fit in a `/16` range.
+The default node allocation of `/24` means a hard cluster node limit of 256 since this is the number of `/24` ranges that fit in a `/16` range.
 
 Depending on the size of the nodes that you are planning to use (in terms of resources), or on the number of nodes you are planning to have,
 the CIDR can be adjusted to be bigger on a per node basis but the cluster would accommodate less nodes overall.
--- a/adoc/deployment-sysreqs.adoc
+++ b/adoc/deployment-sysreqs.adoc
@@ -156,6 +156,30 @@ The management workstation needs at least the following networking permissions:
 * Access to the `apiserver` (the load balancer should expose it, port `6443`), that will in turn talk to any master in the cluster
 * Access to Dex on the configured `NodePort` (the load balancer should expose it, port `32000`) so when the OIDC token has expired, `kubectl` can request a new token using the refresh token
 
+==== Sub-Network Sizing
+
+[IMPORTANT]
+====
+The service subnet and pod subnet must not overlap.
+
+Please plan generously for workload and the expected size of the networks before bootstrapping.
+====
+
+The default pod subnet is `10.244.0.0/16`. It allows for 65536 IP addresses overall.
+Assignment of CIDR's is by default `/24` (256 IP addresses per node) per node.
+
+The default node allocation of `/24` means a hard cluster node limit of 256 since this is the number of `/24` ranges that fit in a `/16` range.
+
+Depending on the size of the nodes that you are planning to use (in terms of resources), or on the number of nodes you are planning to have,
+the CIDR can be adjusted to be bigger on a per node basis but the cluster would accommodate less nodes overall.
+
+If you are planning to use more or less pods per node or have a higher number of nodes, you can adjust these settings to match your requirements.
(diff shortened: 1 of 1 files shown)


adoc/deployment-sysreqs.adoc
```diff
@@ -156,6 +156,30 @@ The management workstation needs at least the following networking permissions:
...
+
+[IMPORTANT]
+====
+The service subnet and pod subnet must not overlap.
+
+Please plan generously for workload and the expected size of the networks before bootstrapping.
+====
+
+The default pod subnet is `10.244.0.0/16`. It allows for 65536 IP addresses overall.
+Assignment of CIDR's is by default `/24` (256 IP addresses per node) per node.
```
Klaven:
```suggestion
Assignment of CIDR's is by default `/24` (256 IP addresses per node).
```

Klaven: 256 I think is a little misleading. hopefully anyone doing this will understand that there are only 254 usable IP's in that range. I might reword this to `(254 usable IP addressed per node)` or something like this as this is what people would really care about anyhow. [1]

adoc/deployment-sysreqs.adoc
```diff
@@ -156,6 +156,30 @@ The management workstation needs at least the following networking permissions:
...
+
+The default node allocation of `/24` means a hard cluster node limit of 256 since this is the number of `/24` ranges that fit in a `/16` range.
```
Klaven: [1] the hard cluster node limit of a `/24` is `254`

adoc/deployment-sysreqs.adoc
```diff
@@ -156,6 +156,30 @@ The management workstation needs at least the following networking permissions:
...
+====
+The service subnet and pod subnet must not overlap.
+
+Please plan generously for workload and the expected size of the networks before bootstrapping.
+====
+
+The default pod subnet is `10.244.0.0/16`. It allows for 65536 IP addresses overall.
+Assignment of CIDR's is by default `/24` (254 usable IP addresses per node).
+
+The default node allocation of `/24` means a hard cluster node limit of 254 since this is the number of `/24` ranges that fit in a `/16` range.
```
Itxaka: Is this a `hard cluster node limit` or a pod limit? Isnt those /24 addresses given to the pods *in* the node, so thats a maximum of 254 pods per node?

* not an expert in networking, disregard and shame me if wrong

adoc/deployment-sysreqs.adoc
```diff
@@ -156,6 +156,30 @@ The management workstation needs at least the following networking permissions:
...
+
+Depending on the size of the nodes that you are planning to use (in terms of resources), or on the number of nodes you are planning to have,
+the CIDR can be adjusted to be bigger on a per node basis but the cluster would accommodate less nodes overall.
+
+If you are planning to use more or less pods per node or have a higher number of nodes, you can adjust these settings to match your requirements.
```
Itxaka: Seems a bit not clear to me. This all means that you can have more pods per node but in return have less nodes overall right?

r0ckarong: Yes, do you have a suggestion on how to make that more clear?

adoc/deployment-sysreqs.adoc
r0ckarong: @ereslibre Could you clarify this for me please.

adoc/deployment-sysreqs.adoc
Itxaka: not really no, sorry.

Maybe merging both statements and clarifying that there is a tradeoff on more pods on less nodes VS less pods per node but more nodes? Not sure how to clearly indicate that, sorry :/

adoc/deployment-sysreqs.adoc
ereslibre: The pod subnet is a cluster-wide limit. Now, if you choose the default: `10.244.0.0/16`, then you can partition that subnet as desired -- the total number of available addresses will be 65536 --. Every node will get assigned a CIDR of its own by Kubernetes. By default it's a `/24` -- check `--node-cidr-mask-size` in the controller manager -- (what allows for 254 IP addresses per node).

Following that reasoning, if every node gets a `/24` when registering, then you can fit 256 nodes in total on a `10.244.0.0/16`.

Depending on the planning of the cluster they might want to change the CIDR allocation for each node (fitting more nodes, less IP addresses per node, or the other way around), or change the global pod subnet altogether.

Issue #598, opened 2019-11-14
[doc] 4.5.2  Deploying a 389 Directory Server with an External Certificate - image 1.4.0 does not have the required and fixed entrypoint script!


Martin-Weiss:
4.5.2  Deploying a 389 Directory Server with an External Certificate

https://susedoc.github.io/doc-caasp/master/caasp-admin/single-html/#_deploying_a_389_directory_server_with_an_external_certificate

------------------------
It seems the whole thing does not work - did anyone else try to work with this??
(just delete and recreate the container with existing data...)

cclhsu:
The command was tested a while back when we have the document https://confluence.suse.com/display/SUSECaaSPlatform4/Setup+389ds+TLS+for+RBAC-Authenticate+with+Self-signed+Certificate.  The docker with new fixed entrypoint is still verifying by QA and has not been released.

Martin-Weiss:
Ok - then we need fixed documentation forthe old image until we have the new image with the fixed entry point released. 

Docs need to match what is released.

r0ckarong:
@cclhsu @innobead Can we provide information on the old image or is the release of the new one imminent?

Martin-Weiss:
> 
> 
> The command was tested a while back when we have the document https://confluence.suse.com/display/SUSECaaSPlatform4/Setup+389ds+TLS+for+RBAC-Authenticate+with+Self-signed+Certificate. The docker with new fixed entrypoint is still verifying by QA and has not been released.

@cclhsu  - where can I find the new entrypoint.sh that is fixed and the steps I need to use with this? Can I use the 1.4.0 image version and just the new entrypoint.sh or are there more changes required to get the SSL certificate replacement working? (I am onsite and have to get this working ASAP).

cclhsu:
Hi! Martin.  The updated entrypoint can be find at https://github.com/SUSE/caasp-container-images/blob/master/caasp-389-ds-image/root/usr/local/bin/entrypoint.sh.  Cert Replacement code is as following:

```
        echo -e "\n>> Deleting ${SERVER_CERT} database in ${SLAP_DIR}...\n"
        # /usr/bin/certuti -d ${SLAP_DIR} -D -n "${SERVER_CERT}" # Remove certificate
        # /usr/bin/certuti -d ${SLAP_DIR} -F -n "${SERVER_CERT}" # Remove key
        echo > ${SLAP_DIR}/pwdfile.txt
        rm -f ${SLAP_DIR}/*.db
[...]

PR #513, opened 2019-09-23
[wip] 389ds changes

cclhsu:
- Instance name hard code to localhost to consist with upstream so user can migrate directory data to newer version.
- Align filter group  to use "groupOfNames" so examples can be used between different sections of documents. 
- make variable consist with changes in 389ds entrypoint.
- Add reference to apply and access dex.yaml for examples

PR #603, opened 2019-11-19
update 389ds external certificate

cclhsu:
# Make sure it's tested
Have used external script to test content of the PR.

# Describe your changes
Update external Certificate replacement steps for old 389ds image.

# Related Issues / Projects
https://github.com/SUSE/doc-caasp/issues/598#issuecomment-554303199

### DIFF
--- a/adoc/admin-security-ldap.adoc
+++ b/adoc/admin-security-ldap.adoc
@@ -14,7 +14,6 @@ with a customized configuration using container commands.
 into the terminal in the following format:
 +
 ----
-INSTANCE_NAME=localhost                         # Directory Server Name
 ROOT_DN="cn=admin,dc=example,dc=org"            # Admin Account
 ROOT_PASSWORD=                                  # Admin Password
 SUFFIX="dc=example,dc=org"                      # Domain Suffix
@@ -29,7 +28,6 @@ automatically self-signed certificate and key.
 docker run -d \
 	-p 389:389 \
 	-p 636:636 \
-	-e INSTANCE_NAME=<INSTANCE_NAME> \
 	-e ROOT_DN=<ROOT_DN> \
 	-e ROOT_PASSWORD=<ROOT_PASSWORD> \
 	-e SUFFIX=<SUFFIX> \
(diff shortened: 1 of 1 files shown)


adoc/admin-security-ldap.adoc
```diff
@@ -47,72 +45,43 @@ To replace the automatically generated certificate with your own, follow these s
...
 . Run bash in entrypoint to access the container with the mounted data directory from the previous step:
 +
 ----
 docker run --rm -it \
-  -p 389:389 \
-  -p 636:636 \
   -v <DATA_DIR>:/data \
   --entrypoint /bin/bash \
-  --name 389-ds registry.suse.com/caasp/v4/389-ds:1.4.0
+  --name 389-ds-cert registry.suse.com/caasp/v4/389-ds:1.4.0
```
Martin-Weiss: I thought we get a new image which includes the fixed entry point? In this case the image version should not be 1.4.0 anymore as this is the broken one.

adoc/admin-security-ldap.adoc
```diff
@@ -29,7 +28,6 @@ automatically self-signed certificate and key.
 docker run -d \
 	-p 389:389 \
 	-p 636:636 \
-	-e INSTANCE_NAME=<INSTANCE_NAME> \
 	-e ROOT_DN=<ROOT_DN> \
 	-e ROOT_PASSWORD=<ROOT_PASSWORD> \
 	-e SUFFIX=<SUFFIX> \
```
Martin-Weiss: When I look at the new entrypoint.sh script at https://raw.githubusercontent.com/SUSE/caasp-container-images/master/caasp-389-ds-image/root/usr/local/bin/entrypoint.sh I see other variable names.

DS_SUFFIX, DOMAIN_SUFFIX and DS_DM_PASSWORD -> so I assume the ROOT_DN, ROOT_PASSWORD and SUFFIX are also not valid anymore?

adoc/admin-security-ldap.adoc
```diff
@@ -29,7 +28,6 @@ automatically self-signed certificate and key.
 docker run -d \
 	-p 389:389 \
```
Martin-Weiss: Just had realized this in the entrypoint.sh:

[slapd]
ServerPort=3389
;SecurePort=3636

--> I had the problem that ldap on 389 was not reachable - so I see that run vs. entrypoint.sh does not match here, too.

Could you either change the run or the entrypoint.sh so that this matches?

Martin-Weiss:
As a general comment - the process with stopping a running container, starting a second one to adjust the data in the first one is not the best process.

IMO we should change this process completely and have only one single run that allows to create a new directory instance if non exists, and / or update the ssl certificates via "delete container, replace files, create container" (update / change certificate) in case directory data already exists.

adoc/admin-security-ldap.adoc
```diff
@@ -47,72 +45,43 @@ To replace the automatically generated certificate with your own, follow these s
 docker stop 389-ds
 ----
 
-. Copy the external certificate `<rootCA.pem>` to a mounted data directory `<DATA_DIR>/config/`.
+. Copy the external private key `Server-Cert.crt`, certificate `Server-Cert.csr`, and password `pwdfile-import.txt` to a mounted data directory `<DATA_DIR>/config/`.
```
Martin-Weiss: I assume Server-Cert.crt is the certificate and not the private key. (crt extension)
I assume Server-Cert.csr is the certificate signing request (csr extension)
-> so are you sure we need crt+crt, here? I would exect to see the crt and key here.

Regarding the pwdfile-import.txt -> what if the key does not have a password set? Is it required to have a password set on the key?

adoc/admin-security-ldap.adoc
cclhsu: This PR is mean to update the document for using old 389ds image.

Martin-Weiss: Ok - could you help me pointing me to the PR for using the new image?

cclhsu:
Since new 389ds image is in progress of verification for 4.0.3 release this sprint, I would close this PR. We should use #513 which have changes to go along with new image.

PR #604, opened 2019-11-19
Unify use of placeholders (Closes #538, Closes #592).

### LINKED ISSUE #538: Unify use of placeholders.
According to the [conventions defined in our guides](https://documentation.suse.com/suse-caasp/4/single-html/caasp-admin/#_documentation_conventions) and the [SUSE Style Guide](http://docserv.nue.suse.com/documents/Style_Guide/docu_styleguide/html/#sec-placeholder), placeholders for actual values are written in capitals: `PLACEHOLDER`. 
In reality we don't really adhere to this rule. The most established way of writing placeholders in our docs is inside angle brackets and lower case: `<placeholder>`. The use of separators in multi word placeholders is also not clear. I've seen periods, underscores, hyphens and spaces. 
Let's decide on a way to do it and stick to it.

### LINKED ISSUE #592: [doc] 5.2  Cluster Deployment -> make clear that "USERNAME" is a variable
5.2  Cluster Deployment

https://susedoc.github.io/doc-caasp/master/caasp-deployment/single-html/#_cluster_deployment

-> echo "USERNAME ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers should be changed to something like

-> echo "\<USERNAME\> ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers

Or could we change USERNAME to a different font?

nkoranova:
Unifies use of placeholders to the <THIS_IS_A_PLACEHOLDER> formatting.

Closes: 
- https://github.com/SUSE/doc-caasp/issues/538 
- https://github.com/SUSE/doc-caasp/issues/592

### DIFF
--- a/adoc/admin-security-ldap.adoc
+++ b/adoc/admin-security-ldap.adoc
@@ -296,7 +296,7 @@ connectors:
     # Please note that if the bind password contains a `$`, it has to be saved in an
     # environment variable which should be given as the value to `bindPW`.
     bindDN: cn=user-admin,ou=Users,dc=example,dc=org
-    bindPW: <BIND DN PASSWORD>
+    bindPW: <BIND_DN_PASSWORD>
 
     # The attribute to display in the provided password prompt. If unset, will
     # display "Username"
--- a/adoc/admin-security-ldap.adoc
+++ b/adoc/admin-security-ldap.adoc
@@ -296,7 +296,7 @@ connectors:
     # Please note that if the bind password contains a `$`, it has to be saved in an
     # environment variable which should be given as the value to `bindPW`.
     bindDN: cn=user-admin,ou=Users,dc=example,dc=org
-    bindPW: <Password of Bind DN>
+    bindPW: <BIND DN PASSWORD>
 
     # The attribute to display in the provided password prompt. If unset, will
     # display "Username"
@@ -305,13 +305,13 @@ connectors:
     # User search maps a username and password entered by a user to a LDAP entry.
     userSearch:
       # BaseDN to start the search from. It will translate to the query
-      # "(&(objectClass=person)(mail=<username>))".
+      # "(&(objectClass=person)(mail=<USERNAME>))".
       baseDN: ou=Users,dc=example,dc=org
       # Optional filter to apply when searching the directory.
       filter: "(objectClass=person)"
 
       # username attribute used for comparing user entries. This will be translated
-      # and combined with the other filter as "(<attr>=<username>)".
+      # and combined with the other filter as "(<attr>=<USERNAME>)".
       username: mail
       # The following three fields are direct mappings of attributes on the user entry.
       # String representation of the user.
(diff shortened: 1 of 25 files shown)


adoc/admin-security-ldap.adoc
```diff
@@ -296,7 +296,7 @@ connectors:
     # Please note that if the bind password contains a `$`, it has to be saved in an
     # environment variable which should be given as the value to `bindPW`.
     bindDN: cn=user-admin,ou=Users,dc=example,dc=org
-    bindPW: <Password of Bind DN>
+    bindPW: <BIND DN PASSWORD>
```
r0ckarong: Here you're missing the underscores.

nkoranova:
This is WIP.  I've gone through all chapters, but I still want to look more closely through `deployment-terraform-example.adoc` and  `deployment-loadbalancer.adoc` as it mentions "Replace the individual `masterXX` with the IP/FQDN of your actual master nodes [...]" and I'm not sure how to handle that yet.

adoc/admin-security-ldap.adoc
r0ckarong: Rest looks good.

nkoranova:
```suggestion
    bindPW: <BIND_DN_PASSWORD>
```

adoc/deployment-loadbalancer.adoc
```diff
@@ -242,7 +242,7 @@ zypper in haproxy
 +
 [NOTE]
 ====
-Replace the individual `masterXX` with the IP/FQDN of your actual master nodes (one entry each) in the `server` lines.
+Replace the individual `<MASTER_XX>` with the IP/FQDN of your actual master nodes (one entry each) in the `server` lines.
```
nkoranova: @r0ckarong what is meant  by "(one entry each)"?  - I'm not sure I understood if both the `<MASTER_00> <MASTER_00>` are supposed to get substituted for the IP/FQDN. Git blame says you wrote this line... :angel:

r0ckarong: For each master node that you want to use you must add another `server ...` line to the nginx.conf file. You replaced it correctly already.

r0ckarong: And only the second one needs to be substituted. The first one is the name the second one will be the URL under which that is reachable.

So `master00 master.example.com:6443` for instance.

nkoranova: By URL you mean FQDN? 
Would this illustrate it more accurately? 
```server <MASTER_00_IP> <MASTER_00_FQDN>:6443 ```

r0ckarong: No I think the first part is an internal name in nginx. The second part must be replaced with EITHER the IP OR FQDN

r0ckarong: @lcavajani Can you clarify what the format for these entries in the config must be?

ereslibre: Each entry corresponds to `server <name> <address>[:[port]] [param*]` ([documentation](https://cbonte.github.io/haproxy-dconv/1.7/configuration.html#server)). So the first argument is the internal `name` of the backend server, used for logging and alerts, whereas the second argument is the address of the backend server.

Issue #607, opened 2019-11-20
Check that our documentation includes NTP client deployment 


nkoranova:
Our documentation has to specify that the bootstrapper node needs to run NTP as well and be synced in terms of time. Bootstrapping skuba using a fully up-to-date system won't be possible unless we have this documented. 

https://bugzilla.suse.com/show_bug.cgi?id=1157063&GoAheadAndLogIn=1

Check that we have this info and include it if we don't.

r0ckarong:
Does this mean, once merged, we have to rebuild the 4.0.2 docs release package?

nkoranova:
@ereslibre how urgent is this? You said it' not critical, but we should have this info public ASAP, right? So rebuilding the 4.0.2 docs package might be a good idea.

ereslibre:
> @ereslibre how urgent is this? You said it's not critical, but we should have this info public ASAP, right? So rebuilding the 4.0.2 docs package might be a good idea.

It's not crit-sit, if that makes sense, but if we have the chance to include this on the 4.0.2 docs I think that would be great. I am struggling with this because I could swear we documented this, but maybe I'm wrong and we never wrote it down on the *public* documentation.

nkoranova:
>  I am struggling with this because I could swear we documented this

Yeah this also has me slightly worried, I really hope it didn't get lost somewhere in our publishing process :-/

r0ckarong:
The terraform examples for both VMWare and SOC mention NTP servers to be used configured in the terraform examples:
https://documentation.suse.com/suse-caasp/4/single-html/caasp-deployment/#_using_terraform
https://documentation.suse.com/suse-caasp/4/single-html/caasp-deployment/#_deploying_the_cluster_nodes

I would have assumed that the clients using these servers are deployed by terraform.

Bare metal needs to be installed via AutoYast and that mentions NTP client as well:
https://documentation.suse.com/suse-caasp/4/single-html/caasp-deployment/#_autoyast_preparation

I currently do not see where this is missing. We do not explicitly state this as a requirement though as I believe we did in v3 docs.

jordimassaguerpla:
Can we do this for 4.0.3? We already released 4.0.2. Unless you consider this a P1 issue. 4.0.3 is already being reviewed and we expect it to be released in the next 2 weeks.

ereslibre:
> I currently do not see where this is missing. We do not explicitly state this as a requirement though as I believe we did in v3 docs.

[...]

PR #608, opened 2019-11-22
Add NTP server on management node note.

nkoranova:
Solves: https://github.com/SUSE/doc-caasp/issues/607

In addition I also slightly changed the heading structure in the next chapter to make it less confusing and more clearly arranged.

### DIFF
--- a/adoc/deployment-sysreqs.adoc
+++ b/adoc/deployment-sysreqs.adoc
@@ -47,15 +47,19 @@ In order to deploy and control a {productname} cluster you will need at least on
 machine capable of running `skuba`. This typically is a regular desktop workstation or laptop
 running SLE 15 SP1 or later.
 
-[NOTE]
-====
 The `skuba` CLI package is available from the {productname} module.
 You will need a valid {sle} and {productname} subscription to install this tool on the workstation.
-====
 
-.Storage Sizing
+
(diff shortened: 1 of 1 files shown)


adoc/deployment-sysreqs.adoc
```diff
@@ -47,25 +47,29 @@ In order to deploy and control a {productname} cluster you will need at least on
...
 The `skuba` CLI package is available from the {productname} module.
 You will need a valid {sle} and {productname} subscription to install this tool on the workstation.
-====
 
-.Storage Sizing
+
+.Time Synchronization
 [IMPORTANT]
 ====
+It is vital that the management workstation runs an NTP server and that time synchronization is configured to the same NTP servers,
```
r0ckarong: I don't understand the line break here in the middle of the sentence. Was that by accident?

r0ckarong: And the wording here is wrong. The workstation doesn't have to provide an NTP server to the cluster nodes. It must be configured so that the local NTP client on the workstation is getting time synchronization from the same servers that will be used for the cluster nodes.

We recommend the customer to provide an NTP server in their network but we don't want to explain how. All we need to convey is that all the NTP clients across the machines must use the same servers in order to get the same timing.

PR #614, opened 2019-11-26
Add PSP creation to prometheus config example

### LINKED ISSUE #484: [doc] 5.1.4  Prometheus
5.1.4  Prometheus

https://www.suse.com/documentation/suse-caasp-3/singlehtml/book-caasp-admin/book-caasp-admin.html#prometheus

Due to upstream changes in the Prometheus Helm Chart, the current steps to deploy Prometheus no longer work and need a slight update.

I was told "This seems to be due to Update psps to support all pods (#15055) · helm/charts@af1e617"
Related: https://github.com/helm/charts/commit/af1e617edb58566f859c8c3aabb509c3edafc85b#r34888663

Per a conversation on the CaaSP mailing list one change could be to add "--set podSecurityPolicy.enabled=true" to the end of the "helm install" line for Prometheus, like this:
helm install --name prometheus stable/prometheus --namespace monitoring --values prometheus-config-values.yaml --set podSecurityPolicy.enabled=true

The better/recommended option would be to add a top level PSP as well:
But for the latest chart version is should add the new top level setting as well:

podSecurityPolicy:
  enabled: true

r0ckarong:
Closes #484

### DIFF
--- a/xml/admin_monitoring.xml
+++ b/xml/admin_monitoring.xml
@@ -415,6 +415,10 @@ Adding password for user admin
       </listitem>
      </itemizedlist>
 <screen language="yaml">
+# Enable the creation of pod security policy
+podSecurityPolicy:
+  enabled: true
+
 # Alertmanager configuration
 alertmanager:
   enabled: true


xml/admin_monitoring.xml
```diff
@@ -415,6 +415,10 @@ Adding password for user admin
       </listitem>
      </itemizedlist>
 <screen language="yaml">
+# Enable the creation of pod security policy
+podSecurityPolicy:
+  enabled: true
```
mjura: I think better would be to change it in our Prometheus Helm charts 
https://github.com/SUSE/kubernetes-charts-suse-com/blob/master/stable/prometheus/values.yaml#L5

mjura: I will take care of it

r0ckarong: This change is for v3 where we use the upstream prometheus chart. Is prometheus available from the suse registry for v3?

mjura: ahhh, I am sorry I didn't notice that it comes from v3. Yes, we are shipping now our own Helm charts and prometheus container in registry. Just wondering if we update documentation for v3 too.

r0ckarong:
> This is CaaSP v3 documentation change and does not affecting v4
> +1

So what is there to do now? Is the change here OK for v3? Or do we need to change the docs to reflect usage of our own image for v3?

xml/admin_monitoring.xml
mjura: I have submitted PR to SUSE Prometheus helm charts https://github.com/SUSE/kubernetes-charts-suse-com/pull/105

mjura:
> > This is CaaSP v3 documentation change and does not affecting v4
> > +1
> 
> So what is there to do now? Is the change here OK for v3? Or do we need to change the docs to reflect usage of our own image for v3?

@kkaempf Could you please answer if we want waste some time on fixing CaaSP v3 documentation about using SUSE Prometheus and Grafana helm charts and images?

kkaempf:
No. CaaSP v3 is in (reactive) maintenance. Bugfixing only.
SUSE-provided prometheus and grafana is v4-only.

PR #618, opened 2019-11-28
Mention other methods for exposing monitoring services

thehejik:
The documentation is still referring to nginx ingress controller and TLS but I think it is worth to mention also other methods for exposing P&G - esp. when the services are not reachable from outside.


jordimassaguerpla:
thanks

PR #637, opened 2019-12-12
Document skuba log verbosity max of 5

jodavis-suse:
In an effort to make skuba consistent with upstream Kubernetes
logging practices, simplify the log level range from 10 to 5.

# Describe your changes
We will be making some further changes to skuba to simplify the number of log levels used.  This documentation is important to correctly represent that effort.

# Related Issues / Projects
Relates to: SUSE/avant-garde#493
See also SUSE/avant-garde#841

### DIFF
--- a/adoc/admin-troubleshooting.adoc
+++ b/adoc/admin-troubleshooting.adoc
@@ -345,12 +345,12 @@ After opening a Service Request (SR), you can upload the TAR archive to Global T
 
 == Debugging Cluster Deployment
 
-If the cluster deployment fails, please re-run the command again with setting verbosity level to 10 `-v=10`.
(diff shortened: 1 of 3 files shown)


r0ckarong:
@jordimassaguerpla When will this change ship?

Martin-Weiss:
Just to be sure to be sure - in case a customer has scripted this already and specified -v10 or -v99 he will still get all debug? So >5 = 5?

jodavis-suse:
> Just to be sure to be sure - in case a customer has scripted this already and specified -v10 or -v99 he will still get all debug? So >5 = 5?

That is correct, based on my testing.  There will be some follow on work to ensure all the messages logged in the skuba code use values 5 or less.  Currently (and we don't intend to change it) you can enter a log verbosity value up to the max integer value and it will give all messages logged at a lower number without producing exceptions.

PR #641, opened 2019-12-13
Fix loglevel example in troubleshooting to 5

r0ckarong:
Prematurely merged #637 
This reintroduces the change.

@jodavis-suse

### DIFF
--- a/adoc/admin-troubleshooting.adoc
+++ b/adoc/admin-troubleshooting.adoc
@@ -345,12 +345,14 @@ After opening a Service Request (SR), you can upload the TAR archive to Global T
 
 == Debugging Cluster Deployment
 
-If the cluster deployment fails, please re-run the command again with setting verbosity level to 10 `-v=10`.
+== Debugging Cluster Deployment
+
+If the cluster deployment fails, please re-run the command again with setting verbosity level to 5 `-v=5`.
 
 For example, if bootstraps the first master node of the cluster fails, re-run the command like
 [source,bash]
 ----
-skuba node bootstrap --user sles --sudo --target <IP/FQDN> <NODE_NAME> -v=10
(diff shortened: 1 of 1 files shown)


r0ckarong:
> Could it be that we have the loglevel also in "node upgrade apply" help text? Did you change this in all places or just in bootstrap?

The bootstrap section is the only place we specifically run this with loglevel 5.

Martin-Weiss:
So in the node upgrade we still support loglevel 10  or is the output of the node upgrade help text wrong in skuba?

r0ckarong:
> So in the node upgrade we still support loglevel 10 or is the output of the node upgrade help text wrong in skuba?

@jodavis-suse Can you answer this?

jodavis-suse:
I'll admit, I haven't read all the documentation yet, so I had to do some searching.  This is the only instance of a reference to -v=10 or verbosity I found in the Admin Guide.
I dug through the code again to be sure, and I think you are referencing the 'node upgrade apply' functionality which is defined in skuba/node/upgrade.go.  That functionality is based on the same Cobra command line arguments as the rest of Skuba, and the verbosity option and help message that are displayed on the command line are common to all Skuba commands and will now output a message indicating the range is from 0 to 5.
You can also note that the code does not enforce the actual value, and a 10 or even 100 verbosity on the command line will have the same effect as a 5.

Martin-Weiss:
Thanks for the details - in case the Skuba code is already changed to 5 in the "node upgrade apply" and all other help texts and the documentation also states 5 I think we got rid if the confusion ;-).

Probably we should add the -v5 also in the documentation section for "node upgrade apply"?

P.S. my current "best" practice is to use "-v999 2>&1|tee <node-name>-<action-name>.log" with skuba to ensure we get all logs ;-) - maybe we should also document the usage of tee to get a logfile?

PR #661, opened 2020-01-13
Admin guide: add Stratos instructions

flavio:
Explain how to install Stratos on top of CaaSP without having to install
any Cloud Foundry component.

Signed-off-by: Flavio Castelli <fcastelli@suse.com>

# Make sure it's tested
*Technical writers will not always be able to verify the implementation!*

I've tested these instructions inside of a test environment, plus they are a write down of a screencast I shared sometimes ago. Some folks from the sales organization tested them as well.

# Describe your changes

This PR introduces a new section inside of the admin/monitoring documentation. The section explains how to install Stratos on top of a CaaSP cluster **without** having to install any Cloud Foundry piece.

Note well: this new section of our manual does **not** explain how to configure Stratos. This is already done inside of the the SUSE Cloud Application Platform.

On the long term we should probably extract Stratos' documentation from the CaaSP manual and from the CAP one and make it live inside of its own document. In the meantime this interim solution has been approved by @qnetter.

# Related Issues / Projects

Relates to:

* https://github.com/SUSE/avant-garde/issues/1213

# Labels

I've added the right labels to the issue.

That's the reasoning I used:
* P1: there's no P1 associated with that
* Blocked/on-hold: everything is ready and available to our customers. This is just making official something that has been circulating among sales and consulting since some time.
* ReleaseNotes: **YES** - we need to update the release notes once this is merged into the docs to ensure all our users are aware of that.
* v3/v4/v4.x: this applies only to the vv+ product
* Needs Review: no review needed

### DIFF
--- a/adoc/admin-stratos-web-console.adoc
+++ b/adoc/admin-stratos-web-console.adoc
@@ -1,10 +1,6 @@
-:toc:
 = Stratos Web Console
 
-[IMPORTANT]
-====
-Stratos web UI is provided as a "tech preview".
-====
+include::common_tech_preview.adoc[]
 
 [NOTE]
 ====
--- a/adoc/admin-stratos-web-console.adoc
+++ b/adoc/admin-stratos-web-console.adoc
@@ -1,8 +1,24 @@
+:toc:
 = Stratos Web Console
 
+[IMPORTANT]
+====
+Stratos web UI is provided as a "tech preview".
+====
+
+[NOTE]
+====
+If you plan to deploy {suse} {cap} on your {productname}
+cluster please skip this section of the documentation and refer
+to the official {suse} {cap} instructions. This will include Stratos.
+
+https://documentation.suse.com/suse-cap/1.5.1/single-html/cap-guides/#cha-cap-depl-caasp
+====
+
+== Introduction
+
 The Stratos user interface (UI) is a modern web-based management application for
-{kube} and for Cloud Foundry distributions based on {kube} like SUSE
-Cloud Application Platform.
+{kube} and for Cloud Foundry distributions based on {kube} like {suse} {cap}.
 
 Stratos provides a graphical management console for both developers and system
 administrators.
@@ -10,14 +26,8 @@ administrators.
 A single Stratos instance can be used to monitor multiple {kube} clusters
 as long as it is granted access to their {kube} API endpoint.
 
-If you plan to deploy SUSE Cloud Application Platform on your {productname}
-cluster please skip this section of the documentation and refer
-to the official SUSE Cloud Application Platform instructions.
-
-== Introduction
-
 This document aims to describe how to install Stratos in a {productname} cluster
-that doesn't plan to run any SUSE Cloud Application Platform components.
+that doesn't plan to run any {suse} {cap} components.
 
 The Stratos stack is deployed using a helm charts and consists of its web
 UI POD and a MariaDB one that is used to store configuration values.
@@ -29,7 +39,7 @@ UI POD and a MariaDB one that is used to store configuration values.
 The deployment of Stratos is performed using a helm chart. Your remote
 administration machine must have Helm installed.
 
-When using a Helm release minor of 3.0, Helm's Tiller component has to be
+When using a Helm release earlier than 3.0, Helm's Tiller component has to be
 installed and active on your {productname} cluster.
 
 === Persistent Storage
@@ -39,25 +49,9 @@ its data.
 
 The cluster must have a {kube} Storage Class defined.
 
-=== Exposing the Web UI
-
-The web interface of Stratos can be exposed either via a Ingress resource or
-by using a Service of type LoadBalancer or even both at the same time.
-
-An Ingress controller must be deployed on the cluster to be able to expose
-the service using an Ingress resource.
-
-The cluster must be deployed on a platform that can handle `LoadBalancer`
-objects and must have the Cloud Provider Integration (CPI) enabled. This
-can be achieved for example when deploying {productname} on top of OpenStack.
-
-=== Securing Stratos
-
-It's highly recommended to secure Stratos' web interface using TLS encryption.
-
-This can be done by creating a TLS certificate for Stratos.
-
 == Installation
+
+=== Adding helm chart repository and default values
 
 . Add SUSE helm charts repository
 +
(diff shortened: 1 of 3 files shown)


adoc/admin-stratos-web-console.adoc
```diff
@@ -0,0 +1,233 @@
...
+      enabled: true
+      tls: |
+        <output of base64 tls.crt>
+      key: |
+        <output of base64 tls.key>
+----
+
+. Change MariaDB password
++
+The helm chart provisions the MariaDB database with a default weak password.
```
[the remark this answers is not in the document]
flavio: The helm chart of Stratos creates a k8s secret for that, behind the scenes.

adoc/admin-stratos-web-console.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+The cluster must be deployed on a platform that can handle `LoadBalancer`
+objects and must have the Cloud Provider Integration (CPI) enabled. This
+can be achieved for example when deploying {productname} on top of OpenStack.
+
+=== Securing Stratos
+
+It's highly recommended to secure Stratos' web interface using TLS encryption.
+
+This can be done creating a TLS certificate for Stratos.
```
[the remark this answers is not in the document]
flavio: RBAC is already handled by the Stratos chart.

adoc/admin-stratos-web-console.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+The web interface of Stratos can be exposed either via a Ingress resource or
+by using a Service of type LoadBalancer or even both at the same time.
+
+An Ingress controller must be deployed on the cluster to be able to expose
+the service using an Ingress resource.
+
+The cluster must be deployed on a platform that can handle `LoadBalancer`
+objects and must have the Cloud Provider Integration (CPI) enabled. This
+can be achieved for example when deploying {productname} on top of OpenStack.
```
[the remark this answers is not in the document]
flavio: We don't have AWS supported right now. Currently the only platform we support with LoadBalancer support is OpenStack.

flavio:
> quick question, do you know if this requires docker capabilities? I think it's important that we keep track as I think in v5 we will switch to being secure by default and this would pose an issue in the future. Not sure if it is worth mentioning but if it did require docker capabilities it might be worth noting?

I didn't look into that deeply, but I don't think they need any special linux capability. They are not doing any weird thing from what I've seen :)

adoc/admin-stratos-web-console.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+== Prerequisites
+
+=== Persistent Storage
+
+The MariaDB instance used by Stratos requires a persistent storage to store
+its data.
+
+The cluster must have a {kube} Storage Class defined.
+
```
suseclee: Without default storage I had pening status on PVCs.
```
kubectl get pvc -n stratos
NAME                                    STATUS    VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
console-mariadb                         Bound     pvc-cedc6eed-d441-482e-82fd-8c30af174e86   1Gi        RWO            nfs-sc         94m
stratos-console-encryption-key-volume   Pending                                                                                       94m
stratos-console-upgrade-volume          Pending

Events:
  Type    Reason         Age                    From                         Message
  ----    ------         ----                   ----                         -------
  Normal  FailedBinding  4m40s (x362 over 94m)  persistentvolume-controller  no persistent volumes available for this claim and no storage class is set
```
After I made SC to default. all the PVC was bounded.
```
kubectl patch storageclass nfs-sc -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"true"}}}'
```
IMO, it is a good idea mentioning making storage class to default.

adoc/admin-stratos-web-console.adoc
```diff
@@ -0,0 +1,233 @@
...
+== Introduction
+
+This document aims to describe how to install Stratos in a {productname} cluster
+that doesn't plan to run any SUSE Cloud Application Platform components.
+
+The Stratos stack is deployed using a helm charts and consists of its web
+UI POD and a MariaDB one that is used to store configuration values.
+
+== Prerequisites
```
c3y1huang: Maybe we could also mention helm and tiller.
Here is what's on CAP doc.
`Helm's Tiller has to be installed and active, with Tiller on the Kubernetes cluster and Helm on your remote administration machine`

adoc/admin-stratos-web-console.adoc
```diff
@@ -0,0 +1,233 @@
...
+`console.service.type` to be `LoadBalancer`.
++
+The `LoadBalancer` resource can be tuned by changing the values of the other
+`loadBalancer*` params specified inside of the `console.service` stanza.
+
+. Expose the web UI using an Ingress
++
+The Ingress resource can be created by setting
+`console.service.ingress.enabled` to be `true`.
++
```
c3y1huang: Might be worthy to mention ingress.host here. With leaving default blank value leads to
```
Error: render error in "console/templates/ingress.yaml": template: console/templates/__helpers.tpl:151:3:
 executing "ingress.host" at <required "Host name is required" $host>: error calling required: Host name
is required
```

adoc/admin-stratos-web-console.adoc
```diff
@@ -0,0 +1,233 @@
...
+
+The MariaDB instance used by Stratos requires a persistent storage to store
+its data.
+
+The cluster must have a {kube} Storage Class defined.
+
+=== Exposing the Web UI
+
+The web interface of Stratos can be exposed either via a Ingress resource or
+by using a Service of type LoadBalancer or even both at the same time.
```
c3y1huang: Should `NodePort` be another option?

flavio: Yes it is, but it's not really flexible... Everybody usually goes with either Ingress or LoadBalancer for common workload deployments

adoc/admin-stratos-web-console.adoc
flavio: The document already has a section about configuring the storage class to use. I extended this section to mention the usage of a `default` storage class. Thanks

adoc/admin-stratos-web-console.adoc
flavio: Done, thanks

adoc/admin-stratos-web-console.adoc
flavio: Good point, I've added that.

flavio:
> Thanks, lgtm. I left some thoughts after following through with the instructions.

Thanks!

> The deployment ECP went through with no errors. But somehow to access ui seems extremely slow (accessible via cable but just spins on wifi). Don't know if the slowness is expected or the network is just bad on my end, doesn't look normal.

That's strange. I think that was due to ECP. I've deployed stratos inside of AWS multiple times, on different clusters, and it always came online quickly.

flavio:
@r0ckarong I think we can merge it. Do you want me to squash all the commits into a single one?

r0ckarong:
> @r0ckarong I think we can merge it. Do you want me to squash all the commits into a single one?

I'll do a squash when I merge it. I need to review it first.

flavio:
> I'll do a squash when I merge it. I need to review it first.

Sounds good to me, thanks

r0ckarong:
@flavio Some aspects of this are pretty vague. The "Securing Stratos" section says:

```
It's highly recommended to secure Stratos' web interface using TLS encryption.

This can be done by creating a TLS certificate for Stratos.
```

OK so now we assume that the user can generate a TLS cert but where does it need to be placed and how is Stratos configured to use that cert?

PR #672, opened 2020-01-17
adoc: document manual reboots

### LINKED ISSUE #653: [doc] 5.3.2  Completely Disabling Reboots - how to do the reboot manually?
5.3.2  Completely Disabling Reboots

https://susedoc.github.io/doc-caasp/master/caasp-admin/single-html/#_completely_disabling_reboots

------
In case the automated drain and reboot is not used and turned off following this method - what are the manual steps required to be followed for a manual reboot? I assume this needs to be documented. (drain, reboot, uncordon and first all masters in sequential order and then all workers in sequential order?)

mssola:
Added step by step documentation on what to do when rebooting nodes manually.

Fixes #653

### DIFF
--- a/adoc/admin-updates.adoc
+++ b/adoc/admin-updates.adoc
@@ -248,15 +248,16 @@ First of all, you will have to `cordon` and link:{kubedoc}tasks/administer-clust
 ----
 kubectl cordon <NODE_ID>
 kubectl drain --force=true \
-  --ignore-daemonsets=false \
+  --ignore-daemonsets=true \
   --delete-local-data=false \ // <1>
   --grace-period 600 \ // <2>
   --timeout=900s \ // <3>
   <NODE_ID>
 ----
-<1> Continues even if there are pods using `emptyDir` (local data that will be deleted when the node is drained).
-<2> Running applications will be notified of termination and given 10 minutes (`600` seconds) to safely store data.
-<3> Draining of the node will fail after 15 minutes (`900` seconds) have elapsed without success.
+<1> Core components like `kured` and `cilium` are running as `DaemonSet` and draining those pods will fail if this is not set to `true`.
+<2> Continues even if there are pods using `emptyDir` (local data that will be deleted when the node is drained; e.g: `metrics-server`).
+<3> Running applications will be notified of termination and given 10 minutes (`600` seconds) to safely store data.
+<4> Draining of the node will fail after 15 minutes (`900` seconds) have elapsed without success.
 
 [IMPORTANT]
 ====
--- a/adoc/admin-updates.adoc
+++ b/adoc/admin-updates.adoc
@@ -241,6 +241,23 @@ kubectl -n kube-system annotate ds kured weave.works/kured-node-lock='{"nodeID":
 
 This command modifies an annotation (`annotate`) on the daemonset (`ds`) named `kured`.
 
+After you have done this, you will have to manage reboots yourself. In order to do this, you will have to follow some steps whenever you want to issue a reboot for a node. First of all, you will have to `cordon` and `drain` the node:
+
+----
+kubectl cordon NODE
+kubectl drain --force=true --delete-local-data=false --ignore-daemonsets=false NODE
+----
+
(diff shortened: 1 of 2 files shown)


adoc/admin-updates.adoc
```diff
@@ -241,6 +241,23 @@ kubectl -n kube-system annotate ds kured weave.works/kured-node-lock='{"nodeID":
 
 This command modifies an annotation (`annotate`) on the daemonset (`ds`) named `kured`.
 
+After you have done this, you will have to manage reboots yourself. In order to do this, you will have to follow some steps whenever you want to issue a reboot for a node. First of all, you will have to `cordon` and `drain` the node:
+
+----
+kubectl cordon NODE
+kubectl drain --force=true --delete-local-data=false --ignore-daemonsets=false NODE
```
lcavajani: I would also add the following options `--grace-period=600 --timeout=900s` (values I use for CAP) with a note explaining that this has to configured in regards of the application running. App with data may require more time to stop successfully.

Without this, draining could hang if a pod can't be terminated.

r0ckarong: So it will grant the app 600s to shut down (no time unit given here?) and 900s (seconds) until it forces the removal, right?

lcavajani: I would also add a note about `--delete-local-data=false`: Continue even if there are pods using emptyDir (local data that will be deleted when
the node is drained).

For example, this is required when a node is running a `metrics-server` pod.

lcavajani: > So it will grant the app 600s to shut down (no time unit given here?) 
yes

> and 900s (seconds) until it forces the removal, right?

No, it will just fail

adoc/admin-updates.adoc
```diff
@@ -241,6 +241,46 @@ kubectl -n kube-system annotate ds kured weave.works/kured-node-lock='{"nodeID":
...
+----
+kubectl cordon <NODE_ID>
+kubectl drain --force=true \
+  --ignore-daemonsets=false \
+  --delete-local-data=false \ // <1>
+  --grace-period 600 \ // <2>
+  --timeout=900s \ // <3>
+  <NODE_ID>
+----
+<1> Continues even if there are pods using `emptyDir` (local data that will be deleted when the node is drained).
```
lcavajani: I would just add a `e.g: metrics-server` since it's a widely used component

adoc/admin-updates.adoc
```diff
@@ -241,6 +241,46 @@ kubectl -n kube-system annotate ds kured weave.works/kured-node-lock='{"nodeID":
...
 This command modifies an annotation (`annotate`) on the daemonset (`ds`) named `kured`.
 
+When automatic reboots are disabled, you will have to manage reboots yourself.
+In order to do this, you will have to follow some steps whenever you want to issue a reboot marker for a node.
+First of all, you will have to `cordon` and link:{kubedoc}tasks/administer-cluster/safely-drain-node/[`drain`] the node:
+
+----
+kubectl cordon <NODE_ID>
+kubectl drain --force=true \
+  --ignore-daemonsets=false \
```
lcavajani: This should be set to `true` since we have core components running as daemonset like `kured` and `cilium`. If set to `false`, draining will fail because it can't by default remove daemonset pods

adoc/admin-updates.adoc
```diff
@@ -241,6 +241,47 @@ kubectl -n kube-system annotate ds kured weave.works/kured-node-lock='{"nodeID":
...
 This command modifies an annotation (`annotate`) on the daemonset (`ds`) named `kured`.
 
+When automatic reboots are disabled, you will have to manage reboots yourself.
+In order to do this, you will have to follow some steps whenever you want to issue a reboot marker for a node.
+First of all, you will have to `cordon` and link:{kubedoc}tasks/administer-cluster/safely-drain-node/[`drain`] the node:
+
+----
+kubectl cordon <NODE_ID>
+kubectl drain --force=true \
+  --ignore-daemonsets=true \
```
mssola: @r0ckarong you need to update this part, then. There is a list of 4 tips below but only three are listed here.

PR #687, opened 2020-01-23
Add documentation on how to manually update a node (Closes #652)

### LINKED ISSUE #652: [doc] 5.3.1  Disabling Automatic Updates - document manual steps if automation is deactivated
5.3.1  Disabling Automatic Updates

https://susedoc.github.io/doc-caasp/master/caasp-admin/single-html/#_disabling_automatic_updates
----

What does a customer have to do to update his cluster "correct" in case he turns of automatic updates following this way? I assume we should document this in addition to the "turn off" function...

mssola:
Fixes #652

### DIFF
--- a/adoc/admin-updates.adoc
+++ b/adoc/admin-updates.adoc
@@ -233,6 +233,12 @@ When this flag is activated no updates are installed at all.
 
 When OS updates are disabled, then you will have to manage OS updates manually. In order to do so, you will have to call `skuba-update` manually on each node.
 
+[WARNING]
+====
+Do not use `zypper up/zypper patch` commands as these do not manage the {kube} annotations used by `kured`.
+If you perform a manual update using these commands you might render your cluster unusable.
+====
+
 After that, rebooting the node will depend on whether you have also disabled reboots or not. If you have disabled reboots for this node, then you will have to follow the instructions as given in <<Completely Disabling Reboots>>. Otherwise, you will have to wait until `kured` performs the reboot of the node
 
 === Completely Disabling Reboots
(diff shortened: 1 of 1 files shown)


mssola:
As you can see, @Martin-Weiss raised some concerns regarding the internal behavior of `kured`, but I think that these concerns have either been expressed already in their properly assigned issues, or that they are out of the scope for this PR. @Martin-Weiss, could you confirm this is the case? Otherwise, what would you want me to further clarify? Thanks!

adoc/admin-updates.adoc
```diff
@@ -231,6 +231,10 @@ It is not required to reload or restart `skuba-update.timer`.
 The `--annotate-only` flag makes the `skuba-update` utility only check if updates are available and annotate the node accordingly.
 When this flag is activated no updates are installed at all.
 
+When OS updates are disabled, then you will have to manage OS updates manually. In order to do so, you will have to call `skuba-update` manually on each node.
+
```
Martin-Weiss: Add : HINT: Do NOT use "zypper up/zypper patch" as this does NOT manage the kubernetes  annotations used by kured!

adoc/admin-updates.adoc
```diff
@@ -231,6 +231,10 @@ It is not required to reload or restart `skuba-update.timer`.
...
+After that, rebooting the node will depend on whether you have also disabled reboots or not. If you have disabled reboots for this node, then you will have to follow the instructions as given in <<Completely Disabling Reboots>>. Otherwise, you will have to wait until `kured` performs the reboot of the node
+
```
Martin-Weiss: Do we have some details on how to adjust kured configuration? I.e. check every 10 Minutes instead of every 60 Minutes? In case no - we should add that - in case yes - we should link this, here.

r0ckarong: @mssola Can you specify more details for this?

PR #691, opened 2020-01-27
Update vmware version in attributes file.

nkoranova:
Based on [VMware documentation](https://www.vmware.com/resources/compatibility/search.php?deviceCategory=software&details=1&operatingSystems=247&productNames=15&page=1&display_interval=10&sortColumn=Partner&sortOrder=Asc&testConfig=16), SLES 15 should be only compatible with VMware ESXi 6.7. Confirmed [with QA](https://chat.suse.de/channel/caasp-qa?msg=amvXnjnzEZsf6yRvT) and PM.

### DIFF
--- a/adoc/attributes.adoc
+++ b/adoc/attributes.adoc
@@ -24,7 +24,7 @@
 // Component Versions
 
 :cilium_version: 1.5.3
-:envoy_version: 
+:envoy_version:
 :crio_version: 1.16.0
 :dex_version: 2.16.0
 :etcd_version: 3.3.15
@@ -32,7 +32,7 @@
 :kube_version: 1.16.2
 :kubedoc: https://v1-16.docs.kubernetes.io/docs/
 :kured_version: 1.2.0-rev4
-:vmware_version: 6.7.0.20000
+:vmware_version: 6.7
 :helm_tiller_version: 2.16.1
 
 // API versions


Martin-Weiss:
> 
> 
> Based on [VMware documentation](https://www.vmware.com/resources/compatibility/search.php?deviceCategory=software&details=1&operatingSystems=247&productNames=15&page=1&display_interval=10&sortColumn=Partner&sortOrder=Asc&testConfig=16), SLES 15 should be only compatible with VMware ESXi 6.7. Confirmed [with QA](https://chat.suse.de/channel/caasp-qa?msg=amvXnjnzEZsf6yRvT) and PM.

I am not sure if you interpreted the VMware documentation correct. SLES 15 is also supported on VMware ESXi 6.5!

grafik

Also see https://partnerweb.vmware.com/programs/guestOS/guest-os-customization-matrix.pdf

nkoranova:
@Martin-Weiss That is true, I did not notice that little plus sign. :) But we support the latest version (6.7), and having 6.7.0.20000 in the docs is too specific and leads to misinterpretation.

Martin-Weiss:
Yes - we have to use the correct version (6.7) number that is also reflected @ VMware ;-).  
Maybe the Patchlevel also matters if https://partnerweb.vmware.com/programs/guestOS/guest-os-customization-matrix.pdf is relevant..

However - IMO we should not care about the VMware version as this is SLES certification anyway and with CaaSP we are just sitting "on top of the OS" and inherit the SLES certification. From a technical point of view I assume that this is only relevant for VSphere storage integration (that we do not have officially, yet) and maybe for terraform integration.

PR #719, opened 2020-02-25
[WIP] Add more terms to the glossary

r0ckarong:
Let's collect abbreviations that come up all the time. This doesn't have to be merged right away.


r0ckarong:
@nkoranova Should we merge this for now and keep adding stuff along the way? I think it's a good collection for now.

nkoranova:
Agreed! :) Merge it.

PR #723, opened 2020-02-28
Change internal link syntax from angle brackets to xref.

nkoranova:
Solves: https://github.com/openSUSE/daps/issues/545


r0ckarong:
I'm fine with this change.

nkoranova:
Turns out this is not necessary.

PR #724, opened 2020-03-02
Update kured version


evrardjp:
There is no documentation change expected for kured, except a
version change.


evrardjp:
@jordimassaguerpla can we go ahead?
@nkoranova is this okay for you?

Issue #735, opened 2020-03-10
Add documentaton about user sessions in kubeconfig files


jvanz:
Recently, one customer faced some difficulties to use the kubeconfig file on multiples machines. Turn out that was the expected behavior and it would be nice to have a docs about how it works to avoid confusion in the future. 

The bug is [bsc#1165701](https://bugzilla.suse.com/show_bug.cgi?id=1165701)

jvanz:
I'm willing to add these docs. But I would like to know where is the best section to add those.

r0ckarong:
@jvanz Since you asked, I think this should go into the Admin Guide as an addition to the security section explaining how these tokens are used and what to consider.

I'm sure Nora will help you with this.

nkoranova:
Yep, we're on it. :)

PR #737, opened 2020-03-10
Warning about multiples kubeconfig with the same user

jvanz:
Recently we have a customer trying to use multiples kubeconfig file with
the same user. This cause issue because after one of the machines with
the files refresh the OIDC token, the rest cannot refresh the token
anymore. This commit adds a warning about this situation.

Issue https://github.com/SUSE/doc-caasp/issues/735

### DIFF
--- a/adoc/admin-security-rbac.adoc
+++ b/adoc/admin-security-rbac.adoc
@@ -172,8 +172,24 @@ this can lead to issues, because only one of the machines will be able to refres
 The user will be able to download multiple kubeconfig files, but they will only work until one of them needs to refresh the session.
 After that, only one machine will work, namely the first machine which refreshed the token.
 
-One possible workaround for this issue is using multiple users for different kubeconfig
+*A workaround for this issue* is using multiple users for different kubeconfig
 files. See how to add more users in <<_sec.admin.security.users.add>>.
+You can also check information about the user and the respective OIDC tokens in the kubeconfig file under the `users` section:
+
+-----
+users:
+- name: myuser
+  user:
+    auth-provider:
+      config:
+        client-id: oidc
+        client-secret: <SECRET>
+        id-token:  <ID_TOKEN>
+        idp-issuer-url: https://<IP>:<PORT>
+        refresh-token: <REFRESH_TOKEN>
+      name: oidc
+-----
+
 
 ==== Access {kube} Resources
 
--- a/adoc/admin-security-rbac.adoc
+++ b/adoc/admin-security-rbac.adoc
@@ -172,6 +172,8 @@ this can lead to issues, because only one of the machines will be able to refres
 The user will be able to download multiple kubeconfig files, but they will only work until one of them needs to refresh the session.
 After that, only one machine will work, namely the first machine which refreshed the token.
 
+One possible workaround for that is using multiple users for different kubeconfig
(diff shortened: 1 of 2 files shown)


nkoranova:
Great! Many thanks for the PR! 
I thought that it would be an overly long warning and that it deserved it's own section. ;) 
Other than that I have two main points I want to ask about: 

1. How do you know when the tokens need to be refreshed? Do you get some sort of notification or where can you check the "expiration date"? Does the user even need to do it, or is it done automatically?
2. What is the solution to this problem? Are you supposed to use multiple users? How do you then keep track of them?

jvanz:
> Great! Many thanks for the PR!
> I thought that it would be an overly long warning and that it deserved it's own section. ;)
> Other than that I have two main points I want to ask about:
> 
>     1. How do you know when the tokens need to be refreshed? Do you get some sort of notification or where can you check the "expiration date"? Does the user even need to do it, or is it done automatically?

It is done automatically. It is transparent for the user. There is no notification.
> 
>     2. What is the solution to this problem? Are you supposed to use multiple users? How do you then keep track of them?

Use multiple users is a possible solution. But I'm not sure if it is the best one and  I'm not aware of other options.

nkoranova:
> Use multiple users is a possible solution. But I'm not sure if it is the best one and I'm not aware of other options.

@innobead could we please get some input from you for this solution? :)

nkoranova:
@jvanz and @innobead Thanks, this looks pretty good and almost ready to merge. I just have one more question. (Sorry if this is getting slightly annoying, I just want to make sure the user gets all the needed info. :))
We have told the user they should create multiple users for different kubeconfig files. They know how to add a new user and how to create a new kubeconfig file. How do they make sure that the kubeconfig file is added under that user?

jvanz:
> We have told the user they should create multiple users for different kubeconfig files. They know how to add a new user and how to create a new kubeconfig file.

I believe so. @innobead [suggested](https://bugzilla.suse.com/show_bug.cgi?id=1165701#c27) that for the customer. Furthermore, our support should help on that process.

 How do they make sure that the kubeconfig file is added under that user?

To download the file the users need to access the gangway web page or use the cli to download the file. Thus, they need to insert the user credentials. But they can check the user in the kubeconfig file as well:

```
[...]
users:
- name: myuser
  user:
    auth-provider:
      config:
        client-id: oidc
        client-secret: <secret>
        id-token:  <something>
        idp-issuer-url: https://<ip>:<port>
        refresh-token: <token>
      name: oidc
[...]
```

nkoranova:
Awesome, thanks. This is now ready to be merged.

r0ckarong:
> Awesome, thanks. This is now ready to be merged. @r0ckarong Is there a reason not to merge it and publish this in 4.1.2?

If this is not tied to any particular version of dex or gangway I don't see why not.

@jordimassaguerpla Do you have any objections to merging this?

jordimassaguerpla:
> > Awesome, thanks. This is now ready to be merged. @r0ckarong Is there a reason not to merge it and publish this in 4.1.2?
> 
> If this is not tied to any particular version of dex or gangway I don't see why not.
> 
> @jordimassaguerpla Do you have any objections to merging this?

no objection.

PR #760, opened 2020-04-10
Openssh note

### LINKED ISSUE #553: Inaccurate reference to using .ssh/config with skuba
Sorry I didn't see the review request earlier. Unfortunately, skuba uses a golang ssh library which does not check the OpenSSH configuration files. So, while we can set OpenSSH to ignore an identity used by skuba, we can not tell skuba to use or ignore any particular identity. This is unfortunate, as it would be nice to be able to specify a user, identity file, etc in the config for skuba. :/

_Originally posted by @dannysauer in https://github.com/SUSE/doc-caasp/pull/522/review_comment/create_

dannysauer:
# Describe your changes

This change adds a little clarity to the need for an an ssh agent, suggests using newer Eliptic Curve crypto, explains a bit more where keys are used, and removes inaccurate information about the openssh config files (and what the agent caches).

I wrote this months ago, and completely forgot to actually open the PR. :/

# Related Issues / Projects

Closes #553 

This is not release-specific, but probably also isn't critical enough to justify backporting to prior releases - I'd likely just include it going forward.

### DIFF
--- a/adoc/deployment-preparation.adoc
+++ b/adoc/deployment-preparation.adoc
@@ -30,10 +30,13 @@ loaded into the SSH agent.
 ssh-keygen -t ed25519
 ---
 
-#In a later deployment step,
-#skuba will ensure that the key is distributed across all the nodes and trusted by them.
-#For now, you only need to make sure that an ssh-agent is running and that it has
-#the SSH key added:
+////
+// is this still accurate?
+In a later deployment step,
+skuba will ensure that the key is distributed across all the nodes and trusted by them.
+For now, you only need to make sure that an ssh-agent is running and that it has
+the SSH key added:
+////
 
 . The `ssh-agent` or a compatible program is sometimes started automatically by graphical desktop
 environments. If that is not your situation, run:
--- a/adoc/deployment-preparation.adoc
+++ b/adoc/deployment-preparation.adoc
@@ -17,7 +17,7 @@ The use of `ssh-agent` comes with some implications for security that you should
 link:http://rabexc.org/posts/pitfalls-of-ssh-agents[The pitfalls of using ssh-agent]
 
 To avoid these risks please make sure to either use `ssh-agent -t <TIMEOUT>` and specify a time
-after which the agent will self-terminate or terminate the agent yourself before logging out by running `ssh-agent -k`.
+after which the agent will self-terminate, or terminate the agent yourself before logging out by running `ssh-agent -k`.
 ====
 
 To log in to the created cluster nodes from the Management machine, you need to configure an SSH key pair.
--- a/adoc/deployment-preparation.adoc
+++ b/adoc/deployment-preparation.adoc
@@ -19,15 +19,24 @@ after which the agent will self-terminate or terminate the agent yourself before
 `ssh-agent -k`.
 ====
 
-To log in to the created cluster nodes, you need to configure an SSH key pair and
-load it into your user's `ssh-agent` program. This is also mandatory in order to be able to use
-the installation tools `terraform` and `skuba`. In a later deployment step,
-skuba will ensure that the key is distributed across all the nodes and trusted by them.
-For now, you only need to make sure that an ssh-agent is running and that it has
-the SSH key added:
+To log in to the created cluster nodes from the Management machine, you need to configure an SSH key pair.
+This key pair needs to be trusted by the user you will log in as upon each cluster node; this user defaults to
+"sles".  In order to use the installation tools `terraform` and `skuba`, this trusted keypair also must be
+loaded into the SSH agent.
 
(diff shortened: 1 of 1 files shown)


adoc/deployment-preparation.adoc
```diff
@@ -19,15 +19,27 @@ after which the agent will self-terminate or terminate the agent yourself before
...
+loaded into the SSH agent.
+
+. If you do not have an existing ssh keypair to use, run:
++
+---
+ssh-keygen -t ed25519
+---
+
+////
+// is this still accurate?
```
nkoranova, line 34: No idea, let's ask around. :)

nkoranova:
I made some edits based on the RC conversation following this comment: https://chat.suse.de/channel/caasp-qa?msg=bx3TzPkazyFBwjRSs 
@dannysauer please check/test this.

nkoranova:
@dannysauer I need you answer on this, it's been hanging here embarrassingly long.

dannysauer:
> @dannysauer I need you answer on this, it's been hanging here embarrassingly long.

Sorry - I had some notes on your changes, and got distracted while looking for them.  I'll get on this in the morning when I'm not half-asleep. :)

adoc/deployment-preparation.adoc
```diff
@@ -8,42 +8,35 @@ and must be maintained on this machine, but it is not a member of the skuba clus
 [[ssh.configuration]]
 === Basic SSH Key Configuration
 
+*In order to successfully deploy {productname}, you need to have an SSH agent.* This is important, because it is required in order to use the installation tools `skuba` and `terraform`.
```
dannysauer, line 11:
```suggestion
*In order to successfully deploy {productname}, you need to have SSH keys loaded into an SSH agent.* This is important, because it is required in order to use the installation tools `skuba` and `terraform`.
```

adoc/deployment-preparation.adoc
```diff
@@ -8,42 +8,35 @@ and must be maintained on this machine, but it is not a member of the skuba clus
...
 [NOTE]
 ====
 The use of `ssh-agent` comes with some implications for security that you should take into consideration.
 
 link:http://rabexc.org/posts/pitfalls-of-ssh-agents[The pitfalls of using ssh-agent]
 
 To avoid these risks please make sure to either use `ssh-agent -t <TIMEOUT>` and specify a time
-after which the agent will self-terminate or terminate the agent yourself before logging out by running
-`ssh-agent -k`.
+after which the agent will self-terminate or terminate the agent yourself before logging out by running `ssh-agent -k`.
```
dannysauer, line 20:
```suggestion
after which the agent will self-terminate or terminate the agent yourself before logging out by running `ssh-agent -k`.,
```
My bad - I left out a comma which I think needs to be in there to indicate where the clauses are separated.

adoc/deployment-preparation.adoc
```diff
@@ -8,42 +8,35 @@ and must be maintained on this machine, but it is not a member of the skuba clus
...
 ====
 
 To log in to the created cluster nodes from the Management machine, you need to configure an SSH key pair.
-This key pair needs to be trusted by the user you will log in as upon each cluster node; this user defaults to
-"sles".  In order to use the installation tools `terraform` and `skuba`, this trusted keypair also must be
-loaded into the SSH agent.
+This key pair needs to be trusted by the user you will log in as into each cluster node; this user defaults to "sles".
```
dannysauer, line 24:
```suggestion
This keypair needs to be trusted by the account you will log in with on each cluster node; this defaults to "sles".
```
Both yours and mine are a bit awkward.  How about this one? :D 

Related - do we use "account" or "user" or "username"?  Probably oughtta be consistent. :)

dannysauer:
Took longer than planned.  I had a couple further minor suggestions.  I'll leave it up to you to decide if they're improvements or not, @nkoranova :)  I'm happy with it when you are.

adoc/deployment-preparation.adoc
nkoranova: I'm pretty sure we mostly use "user", but I like "user account", so I would go with that :D and I also did a search for "key pair" and "keypair", it seems that the former us used more frequently, "keypair" is actually identified as a spelling mistake by my spellchecker.

nkoranova:
I'm happy with this, let's merge it. :fireworks: :confetti_ball:

adoc/deployment-preparation.adoc
dannysauer: "Key Pair" as two words is how the openssh man page describes it, but usually they just say "key." So that'll be my answer as well going forward. :)  Good catch; referring to the pair is an an old habit of mine I should probably break.

PR #770, opened 2020-04-20
Add networking whitelist to deployment requirements (Closes #600)

### LINKED ISSUE #600: [doc] 4.1.3  Installation Tools - proxy server whitelist missing
4.1.3  Installation Tools

https://susedoc.github.io/doc-caasp/master/caasp-deployment/single-html/#_installation_tools

----------------
When deploying CaaSP behind a proxy server we need to have the whitelist of remote URLs that need to be allowed in the proxy.

Can we get the list of URLs / Servers required for SMT/RMT, for HELM and Images that a customer needs to allow in his proxy for CaaSP?

r0ckarong:
Adds a list of resources that need to be reachable (or mirrored) for CaaS Platform.

### DIFF
--- a/adoc/deployment-sysreqs.adoc
+++ b/adoc/deployment-sysreqs.adoc
@@ -355,6 +355,43 @@ net.ipv4.ip_forward = 1
 
 `net.ipv4.ip_forward` must be set to `1`. Additionally, you can check in what order persisted rules are processed by running `sysctl --system -a`.
 
+==== Networking Whitelist
+
+Besides the {suse} provided packages and containers, {productname} uses a couple of elements that are provided by third parties and must be downloaded by {kube}.
+
+The following resources must be available:
+
+{suse} resources:
+
+|===
+|URL|Name|Purpose
+
+|scc.suse.com|{scc}|Allow registration and license activation
+|registry.suse.com|{suse} container registry|Provide container images
+|kubernetes-charts.suse.com|{suse} helm charts repository|Provide helm charts
+|updates.suse.com|{suse} package update channel|Provide package updates
+|===
+
+Upstream / Third-Party resources
+
+|===
+|URL|Name|Purpose
+
+|k8s.gcr.io|Google Container Registry|Provide container images
+|*.cloudfront.net|Cloudfront|CDN/distribution backend for `registry.suse.com`
+|kubernetes-charts.storage.googleapis.com|Google Helm charts repository|Provide helm charts
+|docker.io|Docker Container Registry|Provide container images
+|quay.io|RedHat Container Registry|Provide container images
+|===
+
+Please note that not all installation scenarios will need all of these resources.
+
+[NOTE]
+====
+If you are deploying into an air gap scenario, you must ensure that the resources required
+from these locations are present and available on your internal mirror server.
+====
+
 ==== Communication
 
 Please make sure that all your Kubernetes components can communicate with each other.
--- a/adoc/deployment-sysreqs.adoc
+++ b/adoc/deployment-sysreqs.adoc
@@ -357,28 +357,28 @@ net.ipv4.ip_forward = 1
 
 ==== Networking Whitelist
 
-Besides the {suse} provided packages and containers, {productname} uses a couple of elements that are provided by third parties and must be downloaded by {kube}.
+Besides the {suse} provided packages and containers, {productname} is typically used with third party provided containers and charts.
 
-The following resources must be available:
+The following {suse} provided resources must be available:
 
-{suse} resources:
-
+[options="header,autowidth"]
 |===
 |URL|Name|Purpose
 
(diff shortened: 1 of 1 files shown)


kkaempf:
Hmm, this sounds as if CaaSP wouldn't run without 'external images' ?!

@ereslibre @flavio - this needs your review

lcavajani:
@r0ckarong maybe just mark `Upstream / Third-Party resources` as optional and specify that these resources are used very often in the Kubernetes ecosystem.

We could also add a column to each tab with the equivalent in airgap environment to make it clear, wyt ?

r0ckarong:
> @r0ckarong maybe just mark `Upstream / Third-Party resources` as optional and specify that these resources are used very often in the Kubernetes ecosystem.
> 
> We could also add a column to each tab with the equivalent in airgap environment to make it clear, wyt ? It might be a bit redundant for each line but at least this would clear.

I'm not sure I understand what that would look like, could you give me an example?

r0ckarong:
@kkaempf Still waiting for review by @ereslibre and @flavio from my side this is ready to be merged. Add blocked label again.

r0ckarong:
@evrardjp

adoc/deployment-sysreqs.adoc
```diff
@@ -355,6 +355,43 @@ net.ipv4.ip_forward = 1
...
+|URL|Name|Purpose
+
+|scc.suse.com|{scc}|Allow registration and license activation
+|registry.suse.com|{suse} container registry|Provide container images
+|*.cloudfront.net|Cloudfront|CDN/distribution backend for `registry.suse.com`
+|kubernetes-charts.suse.com|{suse} helm charts repository|Provide helm charts
+|updates.suse.com|{suse} package update channel|Provide package updates
+|===
+
+For additional Upstream / Third-Party resources please also allow the following:
```
evrardjp, line 375: These are not necessary, even in Airgapped environments, for a clean deployment. Of course, it doesn't mean that much can be done afterwards if the most common helm charts and registries aren't available...

I would rephrase this as:

```
It would also be convenient in your kubernetes usage, to allow network access to following resources:
```

adoc/deployment-sysreqs.adoc
```diff
@@ -355,6 +355,43 @@ net.ipv4.ip_forward = 1
...
+|kubernetes-charts.storage.googleapis.com|Google Helm charts repository|Provide helm charts
+|docker.io|Docker Container Registry|Provide container images
+|quay.io|RedHat Container Registry|Provide container images
+|===
+
+Please note that not all installation scenarios will need all of these resources.
+
+[NOTE]
+====
+If you are deploying into an air gap scenario, you must ensure that the resources required
```
evrardjp, line 391: I am not sure it's necessary to have quay/dockerhub/gcr images/... . I think we should not encourage in airgap to allow all of those. I would rather that it's blocked and we get a bug, then a security issue because someone pulled something silly from dockerhub :p

r0ckarong: > I would rather that it's blocked and we get a bug, 

That's the problem, this is not how it works. They will test, will not be able to do what they need with our docs and just not buy the product because someone else explained it better or made it possible.

adoc/deployment-sysreqs.adoc
r0ckarong: I don't understand this view. A "clean" deployment contains no workloads, where would they come from? Certainly not from SUSE because we sell a platform that you can run "whatever" on. If they are in an air gap environment then our main job (besides providing a stable platform and how to set it up) is to explain how to make workloads function. This absolutely requires external charts, images etc.

lcavajani: I think we're not yet talking about airgap here, these are just required if you have an internal cluster which is not meant to be air-gapped from Internet but where you still need to filter what is authorized.

lcavajani: If you run an airgrap cluster, then you have absolutely no direct access to these third party resources so in that case, it's the job of the cluster administrator to set up internal mirror registries.

But in general, the way I see it for the majority it that, when you run an air-grapped cluster, you don't want to fetch anything from these third party resources and you use only internal stuff.

adoc/deployment-sysreqs.adoc
lcavajani: it does not apply here, when you deploy in air gap  scenario, you don't need access to these resources. I would just add, `If you need anything from these location, ensure that the resources required from these locations are present and available on your internal mirror server.`

evrardjp: Yes that's even better wording.

adoc/deployment-sysreqs.adoc
evrardjp: Agreed.

PR #682, opened 2020-01-22
Create document for master nodes disaster recovery

c3y1huang:
# Make sure it's tested
This is tested locally with libvirt. 
QA e2e testcase [PR](https://github.com/SUSE/caasp-test-cases/pull/17) to be verified.

# Describe your changes
This is new commits for disaster recovery.
Includes:
* backup kubernetes data and etcd cluster data.
* restore master nodes.

# Related Issues / Projects
Relates to:
* https://github.com/SUSE/avant-garde/issues/1184

# Enable maintainer updates
Please enable maintainer updates so we can push commits into your branch to make collaboration and reviews easier.

# Do not force push your branch
Please avoid force pushing to branches that are subject of pull requests. Force pushing breaks maintainer commits in many cases and is very hard (if not impossible) to untangle for backporting.

# Labels
Please set any (and all) appropriate labels that describe the status of the PR.

| **Label** | **Description** |
| --- | --- |
| P1 | PR should be worked on and merged as soon as possible |
| Blocked | Work can not proceed because other work has not been completed, PR can not be merged (code has not been merged but documentation is ready) |
| On-Hold | Underlying work is completed but the PR should not be merged |
| ReleaseNotes | User interaction is required after the introduction of this change and the change must be mentioned in the release notes |
| v3/v4/v4.x | Which version of the release the PR should be merged into, this can be multiple versions, please set the "Backport" label if it needs to go into a previous release |
| Needs Review | Some details of the PR are known to be incomplete and must be discussed with other engineers before merging (if possible assign reviewers or cc mention in comments), PR can not be merged |

Signed-off-by: Chin-Ya Huang <chin-ya.huang@suse.com>

PR #791, opened 2020-05-06
Disaster Recovery Chapter Reorganization

nkoranova:
Based on [previous discussion](https://github.com/SUSE/doc-caasp/pull/682#issuecomment-614150470 ), the need arose to clearly distinguish between disaster recovery using Velero and cluster disaster recovery using etcd. The proposed solution is to split the Disaster Recovery chapter in the Admin Guide into two separate chapters - one of which is solely about Velero and how to use it for backing up and restoring resources and one about cluster (specifically master node) disaster recovery using etcd.

The relevant part of the table of contents would look like this: 
image


innobead:
LGTM. 

@jenting @c3y1huang @cclhsu @maximenoel8

nkoranova:
> I fixed the indentation of the prerequisites chapter, rest looks fine.
thanks!
> Please have another look that it's correct the way you wanted it and then feel free to merge.
I think it looks even better than I wanted it to. ;) 
I'll go and merge this now. But I will also create another issue, because now I look at it I think we are missing a short intro to etcd - what is cluster backup good for, what is etcd, how does it differ from Velero and role it plays in disaster recovery. It doesn't have to be as comprehensive as Velero, but we should give the chapter at least a basic intro.

PR #806, opened 2020-05-15
Remove reference to crio.conf


evrardjp:
Starting with crio 1.18, crio.conf will be moved towards
a crio.conf.d folder. All references to crio.conf should be
removed from 4.3.0 docs.

Related: https://github.com/SUSE/avant-garde/issues/1420


evrardjp:
On hold until 4.3.0 is out.

PR #807, opened 2020-05-15
Fix airgapped mirror docs


evrardjp:
This image slipped through the cracks. Hyperkube needs to be mirrored
for old clusters or clusters currently migrating, and starting
with 4.3.0, the image won't be necessary anymore.

This should help on the airgapped story.

Related: https://github.com/SUSE/avant-garde/issues/1420


evrardjp:
on hold until 4.3.0 is out.

PR #809, opened 2020-05-18
Fix numbered list and indentation

thehejik:
Found one issue with numbered list (should be solved by `[start=4]`) and indentation (should be solved by a newline) when after using `include::<file>` the list numbering is reset.

I didn't try to generate any output so it's untested - I just tried adoc WYSIWYG preview in vscode.


nkoranova:
Fixed conflict locally and merged directly. Thanks for the PR!

PR #816, opened 2020-05-22
Add <productname>/<productnumber> tags

sknorr:
Without productname/productnumber tags, most HTML output files won't have a good on-page
indication what version of CaaSP is documented.

If you agree to this, backport to 4.x might be helpful.

### DIFF
--- a/adoc/docinfo.xml
+++ b/adoc/docinfo.xml
@@ -6,3 +6,6 @@
    <dm:editurl>https://github.com/SUSE/doc-caasp/edit/master/adoc/</dm:editurl>
    <dm:translation>yes</dm:translation>
 </dm:docmanager>
+
+<productname>SUSE CaaS Platform</productname>
+<productnumber>5.0.0</productnumber>


nkoranova:
Without feature/with feature:
image
I like it. :)

sknorr:
Unfortunately, yes. The feature currently only works from the included docinfo.xml, and since you can't use AsciiDoc attribute entries in XML files, you can't automatically adjust this.

Not sure how hard it would to expose this directly in AsciiDoc and massage this the postprocess stylesheet. But that's probably possible.

nkoranova:
Phew, ok. I'm not too keen on adding another manual step to every release, since we have those (ideally) every two weeks. We usually just adjust the version number through our adoc entities, this feels like something we will easily forget to do. It does look nicer, but not sure it's worth it. I'm going to wait for what @r0ckarong has to say about it.

sknorr:
You could use a script. E.g., I'd be happy to lend you a version of this if you like:

* script: https://github.com/openSUSE/suse-xsl/blob/master/versionbump
* config: https://github.com/openSUSE/suse-xsl/blob/master/.versionrc

(Notably, the front page of the documentation is not the only spot where you don't have any version information -- in most other SUSE documentation, every page has this info on it: http://documentation.suse.com/ses/5.5/html/ses-all/storage-salt-cluster.html -> "Applies to ..." at the top. In the future stylesheets, I'll try to make sure that this becomes more prominent and is permanently visible on-screen, instead of just at the top of the page.)

r0ckarong:
> The feature currently only works from the included docinfo.xml, and since you can't use AsciiDoc attribute entries in XML files, you can't automatically adjust this.

We're working towards eliminating updating versions manually altogether at the moment. I've added the version number for 5.0 with this commit: baab1bb

And I will ask to have the version update script update this file as well.

sknorr:
Thanks!

PR #866, opened 2020-06-04
Remove idprefix

fsundermeyer:
NO QA required!

Currently all IDs are automatically prefixed with an underscore "_", This results in

- HTML file names starting with an underscrore
- URLs containing lots of underscores

First looks awkward, second earns us SEO penalities from Google

This PR fixes that.
NOTE: HTML file names will be different after merging this PR. 
In  order to not break deep links from outside into old versions I suggest to only apply this to future versions of the CaaSP doc.

In addition to this change I removed underscores and "." from IDs and replaced them with "-". (This was requested for all of doc some time ago, I assume I just found a few IDs that have been overlooked).

I also took the liberty to change the ID syntax from

`[[ID]]`to `[#ID]`

because the latter syntax can also be used inline with asciidoctor
and therefore this change guarantees that IDs are defined in a unique fashion throughout the document.


r0ckarong:
Removed some false positives that identified parts of the CRIO config files. Other than that looks fine. No backport to 4.x

r0ckarong:
@nkoranova @jordimassaguerpla This will break deep links into v5 docs. We need to check all links for v5 in RN and other relevant places.

PR #874, opened 2020-06-09
Correct kustomize dex example

### LINKED ISSUE #808: [doc] 4.6.3.1.2  Example 2: Dex LDAP TLS Connector Configuration (<code class="literal">addons/dex/patches/custom.yaml</code>) -> need correct kustomization and filename
4.6.3.1.2  Example 2: Dex LDAP TLS Connector Configuration (<code class="literal">addons/dex/patches/custom.yaml</code>)

https://susedoc.github.io/doc-caasp/master/single-html/caasp-admin/#_example_2_dex_ldap_tls_connector_configuration_addonsdexpatchescustom_yaml

----------------------------------------------

1. the filename addons/dex/patches/custom.yaml is different than ~/clusters/<CLUSTER_NAME>/addons/dex/patches/dex-patch.yaml a  few chapters further down.

2. the documented example is not a valid dex-patch.yaml as far as I understand and the kubectl apply -k will fail.

--> could we standardize the patch file names and give valid examples how the files should look like "exactly"?

jenting:
# Describe your changes

1. Aligns patch file to `custom.yaml`
2. Corrects file extension from yml to yaml
3. Fix the dex custom.yaml patch example

# Related Issues / Projects

fix #808

### DIFF
--- a/adoc/admin-security-rbac.adoc
+++ b/adoc/admin-security-rbac.adoc
@@ -98,7 +98,7 @@ It is a known limitation of the addon customization process.
 Administrators can update the authentication connector settings after {productname}
 deployment as follows:
 
-. Based on the manifest in `my-cluster/addons/dex/base/dex.yaml`, provide a kustomize patch to `my-cluster/addons/dex/patches/dex-patch.yaml` of the form of strategic merge patch or a JSON 6902 patch.
+. Based on the manifest in `my-cluster/addons/dex/base/dex.yaml`, provide a kustomize patch to `my-cluster/addons/dex/patches/custom.yaml` of the form of strategic merge patch or a JSON 6902 patch.
 +
 Read https://github.com/kubernetes-sigs/kustomize/blob/master/docs/glossary.md#patchstrategicmerge and https://github.com/kubernetes-sigs/kustomize/blob/master/docs/glossary.md#patchjson6902 to get more information.
 +
@@ -152,7 +152,7 @@ kind: Kustomization
 resources:
   - base/dex.yaml
 patches:
-  - patches/dex-patch.yaml
+  - patches/custom.yaml
 ----
 . Apply the changes with:
 +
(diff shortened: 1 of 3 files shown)


adoc/admin-security-rbac.adoc
```diff
@@ -98,7 +98,7 @@ It is a known limitation of the addon customization process.
 Administrators can update the authentication connector settings after {productname}
 deployment as follows:
 
-. Based on the manifest in `~/clusters/<CLUSTER_NAME>/addons/dex/base/dex.yaml`, provide a kustomize patch to `~/clusters/<CLUSTER_NAME>/addons/dex/patches/dex-patch.yaml` of the form of strategic merge patch or a JSON 6902 patch.
+. Based on the manifest in `my-cluster/addons/dex/base/dex.yaml`, provide a kustomize patch to `my-cluster/addons/dex/patches/custom.yaml` of the form of strategic merge patch or a JSON 6902 patch.
```
cclhsu, line 101: `my-cluster` here is a little confusing. I thought it is a configurable 'cluster name'

jenting: All the doc use `my-cluster`. But I agreed it's better to use `<CLUSTER_NAME>`.
WDYT? @r0ckarong @nkoranova

r0ckarong: I need to check if this works properly. Our converter has some issues with using <> in some content types and maybe this needs to be escaped. Otherwise I'm OK with this change.

innobead:
cc @cclhsu @c3y1huang

jenting:
Merge remote-tracking branch 'upstream/maintenance/CaaS4'

adoc/admin-security-rbac.adoc
nkoranova: I tried replacing all "my-cluster" mentions with <CLUSTER_NAME> and I didn't see a converter issue, but to me it feels slightly strange in come contexts. For example when my-cluster is the actual directory of the cluster and we tell the user to go into it: 
```
cd <CLUSTER_NAME> 
```
I think this requires a separate issue and PR.

innobead:
@nkoranova please help also make sure the change into the master branch (it should be the latest updated doc). Thanks.

PR #879, opened 2020-06-11
Add kubernete control plane certificate auto rotation

jenting:
# Description
Add kucero (kubernete control plane certificate rotation) addon which handles
- the administrator is not willing to upgrade the cluster, the certificate might expire after 1 year.
- the kubelet server certificate auto rotation.

# Related Issues / Projects

Relates to: https://github.com/SUSE/avant-garde/issues/1632


nkoranova:
@jenting as I understand it `kucero` is a new function and only relevant for v5. Or should we also backport this to v4?

jenting:
@nkoranova you are right, kucero is a new feature to 5.0

Issue #885, opened 2020-06-12
[doc] 4.10.4.1  Prerequisites - message: secret/oidc-gangway-cert not labeled


Martin-Weiss:
4.10.4.1  Prerequisites:

https://susedoc.github.io/doc-caasp/master/single-html/caasp-admin/#_prerequisites_2

----------------
```
caaspadm@caasp4-01:~/caasp4/suse.cert-exporter> kubectl label --overwrite secret oidc-dex-cert -n kube-system caasp.suse.com/skuba-addon=true
kubectl label --overwrite secret oidc-gangway-cert -n kube-system caasp.suse.com/skuba-addon=true
kubectl label --overwrite secret metrics-server-cert -n kube-system caasp.suse.com/skuba-addon=true
secret/oidc-dex-cert not labeled
caaspadm@caasp4-01:~/caasp4/suse.cert-exporter> kubectl label --overwrite secret oidc-gangway-cert -n kube-system caasp.suse.com/skuba-addon=true
secret/oidc-gangway-cert not labeled
caaspadm@caasp4-01:~/caasp4/suse.cert-exporter> kubectl label --overwrite secret metrics-server-cert -n kube-system caasp.suse.com/skuba-addon=true
secret/metrics-server-cert not labeled
```

--> why do I get the message "not labeled"?

jenting:
I think it's due to skuba `4.2.1` helps to label the secret for you.

We added the prerequisites to the user is because the cert-exporter is delivered by helm-chart, which means it can be run on all Kubernetes versions we delivered.
If the user deploys a cluster with skuba `4.2.0` and the user want to install cert-exporter to monitors the certificate, then the user must do the prerequisites to let cert-exporter know which certificates are addons.

Martin-Weiss:
Ok - then we should at least inform the admin that he might see the message "not lableled". (getting an error message might make admins nervous ;-))

jenting:
sure~

r0ckarong:
@jenting If you create a PR in response to an issue, please use the "Link Issue/PR" feature in the sidebar. This makes attributing this easier while tracking and also closes the issue when merging PR through the GH interface. Automation keywords are great for closing this on the Git CLI but for project tracking please also link the PR issues.

/cc @innobead

PR #888, opened 2020-06-16
Add note that if labels already existed

jenting:
# Describe your changes
What is this PR about? Please briefly detail why this needs to be integrated
and where you think it should go if it's not obvious from the changes.

# Related Issues / Projects

fixes: #885

### DIFF
--- a/adoc/admin-security-certificates.adoc
+++ b/adoc/admin-security-certificates.adoc
@@ -176,7 +176,7 @@ secret/oidc-dex-cert not labeled
 secret/oidc-gangway-cert not labeled
 secret/metrics-server-cert not labeled
 ----
-This is becuase the {skuba} already adding the labels for you.
+This is because the {skuba} has already added the labels for you.
 ====
 
 === Prometheus Alerts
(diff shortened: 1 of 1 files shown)


adoc/admin-security-certificates.adoc
```diff
@@ -166,6 +166,18 @@ kubectl label --overwrite secret oidc-dex-cert -n kube-system caasp.suse.com/sku
...
+[NOTE]
+====
+You might see the console outputs:
+[source,bash]
+----
+secret/oidc-dex-cert not labeled
+secret/oidc-gangway-cert not labeled
+secret/metrics-server-cert not labeled
+----
+This is becuase the {skuba} already adding the labels for you.
```
innobead, line 179:
```suggestion
This is because the {skuba} has already added the labels for you.
```

nkoranova:
@jenting is this only relevant for v4 or should we also include this in v5? @r0ckarong labeled the original issue as 4.2.1, 5.0, v4, v5 and Backport. So I expect that I should also cherry pick this to v5, but I'm not 100% sure.

jenting:
@nkoranova 4.2.1 & v5

PR #898, opened 2020-06-22
Add guide for monitoring NVIDIA GPUs

cmurphy:
Add a section to the GPU guide for configuring Prometheus and Grafana
monitoring for NVIDIA GPUs.

# Describe your changes
What is this PR about? Please briefly detail why this needs to be integrated
and where you think it should go if it's not obvious from the changes.

# Related Issues / Projects

Relates to: https://github.com/SUSE/avant-garde/issues/1702

### DIFF
--- a/adoc/admin-gpus.adoc
+++ b/adoc/admin-gpus.adoc
@@ -181,3 +181,144 @@ Check the kubelet log on the worker node that has a GPU. This may indicate error
 ----
 # journalctl -u kubelet
 ----
+
+== Monitoring
+
+If you have configured link:{docurl}/single-html/caasp-admin/#_monitoring[Monitoring] for your cluster, you may want to use {nvidia}'s link:https://developer.nvidia.com/dcgm[Data Center GPU Manager (DCGM) to monitor your GPUs]. DCGM integrates with the Prometheus and Grafana services configured for your cluster. Follow the steps below to configure the Prometheus exporter and Grafana dashboard for your {nvidia} GPUs.
+
+=== Configure a Pod Security Policy
+
+The DCGM requires use of the hostPath volume type to access the kubelet socket on the host worker node. Create an appropriate Pod Security Policy and RBAC configuration to allow this:
+
+
+[source,bash]
+----
+$ kubectl apply -f - <<EOF
+---
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: nvidia.dcgm
+spec:
+  privileged: false
+  seLinux:
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  runAsUser:
+    rule: RunAsAny
+  fsGroup:
+    rule: RunAsAny
+  allowedHostPaths:
+    - pathPrefix: /var/lib/kubelet/pod-resources
+  volumes:
+  - hostPath
+  - configMap
+  - secret
+  - emptyDir
+  - downwardAPI
+  - projected
+  - persistentVolumeClaim
+  - nfs
+  - rbd
+  - cephFS
+  - glusterfs
+  - fc
+  - iscsi
+  - cinder
+  - gcePersistentDisk
+  - awsElasticBlockStore
+  - azureDisk
+  - azureFile
+  - vsphereVolume
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: nvidia:dcgm
+rules:
+- apiGroups:
+  - policy
+  resources:
+  - podsecuritypolicies
+  verbs:
+  - use
+  resourceNames:
+  - nvidia.dcgm
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: nvidia:dcgm
+roleRef:
+  kind: ClusterRole
+  name: nvidia:dcgm
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind: Group
+  name: system:serviceaccounts:dcgm
+EOF
+----
+
+=== Create the DCGM Exporter
+
+The DCGM exporter monitors GPUs on each worker node and exposes metrics that can be queried.
+
+[source,bash]
+----
+$ kubectl create namespace dcgm
+$ kubectl create --namespace dcgm -f https://raw.githubusercontent.com/NVIDIA/gpu-monitoring-tools/master/dcgm-exporter.yaml
+----
+
+Check that the metrics are being collected:
+
+[source,bash]
+----
+$ NAME=$(kubectl get pods --namespace dcgm -l "app.kubernetes.io/name=dcgm-exporter" -o "jsonpath={ .items[0].metadata.name}")
+$ kubectl port-forward $NAME 8080:9400
+$ # in another terminal
+$ curl http://127.0.0.1:8080/metrics
+----
+
+=== Configure Prometheus
+
+After deploying Prometheus as explained in link:{docurl}/single-html/caasp-admin/#_monitoring[Monitoring], configure Prometheus to monitor the DCGM pods. Gather the cluster IPs of the pods to monitor:
+
+[source,bash]
+----
+$ kubectl get pods --namespace dcgm -l "app.kubernetes.io/name=dcgm-exporter" -o "jsonpath={ .items[*].status.podIP}"
+10.244.1.10 10.244.2.68
+----
+
+Add the DCGM pods to Prometheus's scrape configuration. Edit the Prometheus configmap:
+
+[source,bash]
+----
+$ kubectl edit --namespace monitoring configmap prometheus-server
+----
+
+Under the ``scrape_configs`` section add a new job, using the pod IPs found above:
+
+[source,yaml]
+----
+scrape_configs:
+...
+- job_name: dcgm
+  static_configs:
+  - targets: ['10.244.1.10:9400', '10.244.2.68:9400']
+...
+----
+
+Prometheus will automatically reload the new configuration.
+
+=== Add the Grafana Dashboard
+
+Import the DCGM Exporter dashboard into Grafana.
+
+In the Grafana web interface, navigate to Manage Dashboards and click `Import`. In the field "Import via grafana.com", enter the dashboard ID `12219`, and click `Load`.
+
+Alternatively, download the link:https://raw.githubusercontent.com/NVIDIA/gpu-monitoring-tools/2.0.0-rc.11/grafana/dcgm-exporter-dashboard.json[dashboard JSON definition], and upload it with the "Upload .json file" button.
+
+On the next page, in the dropdown menu labeled "Prometheus", select "Prometheus" as the data source. Customize the dashboard name and UID to your preference. Then click "Import".
+
+The new dashboard will appear in the Grafana web interface.
--- a/adoc/admin-gpus.adoc
+++ b/adoc/admin-gpus.adoc
@@ -315,10 +315,10 @@ Prometheus will automatically reload the new configuration.
 
 Import the DCGM Exporter dashboard into Grafana.
 
-In the Grafana web interface, navigate to Manage Dashboards and click `Import`. In the field "Import via grafana.com", enter the dashboard ID `12219`, and click `Load`.
+In the Grafana web interface, navigate to menu:Manage Dashboards[] and click menu:Import[]. In the field menu:Import via grafana.com[], enter the dashboard ID `12219`, and click menu:Load[].
 
-Alternatively, download the link:https://raw.githubusercontent.com/NVIDIA/gpu-monitoring-tools/2.0.0-rc.11/grafana/dcgm-exporter-dashboard.json[dashboard JSON definition], and upload it with the "Upload .json file" button.
+Alternatively, download the link:https://raw.githubusercontent.com/NVIDIA/gpu-monitoring-tools/2.0.0-rc.11/grafana/dcgm-exporter-dashboard.json[dashboard JSON definition], and upload it with the menu:Upload .json file[] button.
 
-On the next page, in the dropdown menu labeled "Prometheus", select "Prometheus" as the data source. Customize the dashboard name and UID to your preference. Then click "Import".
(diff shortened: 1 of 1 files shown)


adoc/admin-gpus.adoc
```diff
@@ -181,3 +181,144 @@ Check the kubelet log on the worker node that has a GPU. This may indicate error
...
+...
+- job_name: dcgm
+  static_configs:
+  - targets: ['10.244.1.10:9400', '10.244.2.68:9400']
+...
+----
+
+Prometheus will automatically reload the new configuration.
+
+=== Add the Grafana Dashboard
```
cmurphy, line 313: This doesn't currently work due to https://github.com/SUSE/avant-garde/issues/1720 so this should be considered WIP/on hold/blocked until that is resolved and we can fully verify the steps.

cmurphy:
Updated the grafana section to use the graphical UI to import the dashboard. Using a configmap to create the dashboard doesn't work because the $DS_PROMETHEUS variable is undefined and meant to be filled in by the web interface (see https://github.com/grafana/grafana/issues/10786). Still doesn't fully work due to https://github.com/SUSE/avant-garde/issues/1763.

cmurphy:
Unblocked now that https://github.com/SUSE/avant-garde/issues/1763 is resolved

adoc/admin-gpus.adoc
```diff
@@ -181,3 +181,143 @@ Check the kubelet log on the worker node that has a GPU. This may indicate error
...
+$ kubectl create namespace dcgm
+$ kubectl create --namespace dcgm -f https://raw.githubusercontent.com/NVIDIA/gpu-monitoring-tools/master/dcgm-exporter.yaml
+----
+
+Check that the metrics are being collected:
+
+[source,bash]
+----
+$ NAME=$(kubectl get pods --namespace dcgm -l "app.kubernetes.io/name=dcgm-exporter" -o "jsonpath={ .items[0].metadata.name}")
+$ kubectl port-forward $NAME 8080:9400 &
```
dannysauer, line 278: Is this intentionally backgrounded?  If so, should the job be killed later on, or just left running in the background?  Does it exit after the connection is finished?  I've never used `port-forward`. :)

dannysauer: I do like breaking this up into multiple lines using the shell variable.  :+1:

adoc/admin-gpus.adoc
```diff
@@ -181,3 +181,143 @@ Check the kubelet log on the worker node that has a GPU. This may indicate error
...
+$ curl http://127.0.0.1:8080/metrics
+----
+
+=== Configure Prometheus
+
+After deploying Prometheus as explained in link:{docurl}/single-html/caasp-admin/#_monitoring[Monitoring], configure Prometheus to monitor the DCGM pods. Gather the cluster IPs of the pods to monitor:
```
dannysauer, line 284: Should this link to the multi-page HTML section instead of the single page, so there's less to load when someone clicks?  Is there a better way?  Maybe @r0ckarong or @nkoranova have better ideas; this is a question I regularly have about linking across docs. :)

adoc/admin-gpus.adoc
```diff
@@ -181,3 +181,143 @@ Check the kubelet log on the worker node that has a GPU. This may indicate error
...
+----
+
+Add the DCGM pods to Prometheus's scrape configuration. Edit the Prometheus configmap:
+
+[source,bash]
+----
+$ kubectl edit --namespace monitoring configmap prometheus-server
+----
+
+Under the ``scrape_configs`` section add a new job:
```
dannysauer, line 299: Should this explicitly mention "...using the pod IPs"?  I don't know that I have a strong opinion on that; just a thought.

adoc/admin-gpus.adoc
```diff
@@ -181,3 +181,143 @@ Check the kubelet log on the worker node that has a GPU. This may indicate error
...
+...
+----
+
+Prometheus will automatically reload the new configuration.
+
+=== Add the Grafana Dashboard
+
+Import the DCGM Exporter dashboard into Grafana.
+
+In the Grafana web interface, navigate to Manage Dashboards and click `Import`. In the field "Import via grafana.com", enter the dashboard ID `12219`, and click `Load`.
```
dannysauer, line 317: Should "Manage Dashboards" be formatted specially some way?  I'm not sure offhand how we refer to menu items in our docs.  The button name and menu labels are quoted in the next paragraph, so maybe this should also be quoted?

adoc/admin-gpus.adoc
```diff
@@ -181,3 +181,143 @@ Check the kubelet log on the worker node that has a GPU. This may indicate error
 ----
 # journalctl -u kubelet
 ----
+
+== Monitoring
+
+If you have configured link:{docurl}/single-html/caasp-admin/#_monitoring[Monitoring] for your cluster, you may want to use {nvidia}'s Data Center GPU Manager (DCGM) to monitor your GPUs. Follow the steps below to configure the Prometheus exporter and Grafana dashboard for your {nvidia} GPUs.
```
jodavis-suse, line 187: Do we want to include a link to https://developer.nvidia.com/dcgm for further information?

The text takes a big step here, and as someone unfamiliar with DCGM I had to look up whether Prom and Graf are requirements for DCGM or optional components or something else.  Maybe a line "The data from DCGM can be exported to the Prometheus instance configured for your cluster." to help give a fuller picture.

cmurphy: > Do we want to include a link to https://developer.nvidia.com/dcgm for further information?

Sure

> The text takes a big step here, and as someone unfamiliar with DCGM I had to look up whether Prom and Graf are requirements for DCGM or optional components or something else. Maybe a line "The data from DCGM can be exported to the Prometheus instance configured for your cluster." to help give a fuller picture.

I don't quite follow what's missing here. The line is "If you have configured monitoring..." with a link to how to set up Prometheus and Grafana, "...Follow the steps below to configure the Prometheus exporter and Grafana dashboard". If you have monitoring, you can add this thing to monitor your GPUs. Seems clear to me?

Oh - are you saying that it's not clear that the DCGM *is* the exporter and the dashboard?

adoc/admin-gpus.adoc
```diff
@@ -181,3 +181,143 @@ Check the kubelet log on the worker node that has a GPU. This may indicate error
...
+
+=== Configure a Pod Security Policy
+
+The DCGM requires use of the hostPath volume type to access the kubelet socket on the host worker node. Create an appropriate Pod Security Policy and RBAC configuration to allow this:
+
+
+[source,bash]
+----
+$ cat <<EOF | kubectl apply -f -
```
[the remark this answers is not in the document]
cmurphy: I strive for UUOC wherever possible (jk, will fix)

adoc/admin-gpus.adoc
```diff
@@ -181,3 +181,143 @@ Check the kubelet log on the worker node that has a GPU. This may indicate error
...
+  - cephFS
+  - glusterfs
+  - fc
+  - iscsi
+  - cinder
+  - gcePersistentDisk
+  - awsElasticBlockStore
+  - azureDisk
+  - azureFile
+  - vsphereVolume
```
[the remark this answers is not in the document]
cmurphy: The choice to enumerate the allowed volume types is deliberate, yes some volume types should not be allowed by default and this is why our default PSPs also have an explicit list. This list is that default unprivileged list plus hostPath.

adoc/admin-gpus.adoc
cmurphy: This is copypasted from the upstream documentation, probably to avoid needing to change your terminal after entering the command. It seems to die on its own after a while.

adoc/admin-gpus.adoc
cmurphy: Most of the internal doc references I could find were to the single-html pages. Open to suggestions from the docs team on the preferred way to do this.

adoc/admin-gpus.adoc
cmurphy: This is resolved.

adoc/admin-gpus.adoc
cmurphy: Will await input from the docs team.

adoc/admin-gpus.adoc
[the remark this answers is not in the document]
dannysauer: Good enough for me. :)

adoc/admin-gpus.adoc
cmurphy: Added a sentence that hopefully clarifies a bit

adoc/admin-gpus.adoc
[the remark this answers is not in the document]
cmurphy: done

adoc/admin-gpus.adoc
```diff
@@ -181,3 +181,143 @@ Check the kubelet log on the worker node that has a GPU. This may indicate error
...
+roleRef:
+  kind: ClusterRole
+  name: nvidia:dcgm
+  apiGroup: rbac.authorization.k8s.io
+subjects:
+- kind: Group
+  name: system:serviceaccounts:dcgm
+---
+----
+
```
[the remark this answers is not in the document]
cmurphy: done

adoc/admin-gpus.adoc
jodavis-suse: Yeah, sorry if I wasn't clear (my wordsmithing isn't the best today). I think I was looking for a little bit more detail on the relation between DCGM and Prometheus, and basically I think that is that DCGM includes a metrics exporter which can be configured to work with Prometheus (or apparently some other monitoring systems, but we don't need to mention that).

adoc/admin-gpus.adoc
cmurphy: done

adoc/admin-gpus.adoc
dannysauer: I'm back, and better informed! :D
According to https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#port-forward, the forwarding remains until the connected-to pod terminates.  So this would otherwise hang around until the shell session ends.

As such, it might not be bad to provide a command to stop the backgrounded port forwarding.  Either `kill %kubectl` or `kill %+` would be good options.  Those kill a backgrounded kubectl (if there's only one) or the most recent backgrounded job, respectively - I think I prefer the `%+` here.  Otherwise there's maybe a potential security hole leaving a port open, and there's the annoying "there are stopped jobs" when you try to exit the shell.

I'm not really tied to that needing to be there, though, so it's up to you if adding the cleanup is worth the extra clutter.

cmurphy: I think this is getting too into the weeds, this is just supposed to be a quick example to check the functionality of the metrics server, not a lesson in bash job control. Moreover I think it's dangerous to instruct users to use `kill` when we don't know for sure what other processes they may have stopped in between running these examples. I can remove the `&` if that would put you more at ease?

dannysauer: If you remove the `&`, the forward stays in the foreground, which requires opening another terminal. That seems worse.

It seems pretty unlikely that the user would start the forwarding, do the curl, and then take a moment to restart the enterprise payroll database in the same shell session before finishing the example. But maybe I'm overestimating the target audience. ;)

Like I said, I'm not married to the cleanup command, so if that's concerning, I approve of this as-is as well. 👍

cmurphy: > If you remove the &, the forward stays in the foreground, which requires opening another terminal.

Yes, but then it's deliberate, and then they either ^C the port-forward or close that terminal window when they are done, and your concern seems to be that we're leaving garbage processes running without the user realizing it.

> That seems worse.

Worse from a security perspective, or a user-friendliness perspective?

> It seems pretty unlikely that the user would start the forwarding, do the curl, and then take a moment to restart the enterprise payroll database in the same shell session before finishing the example.

It doesn't have to be that drastic, maybe they start the forwarding, open vim to look up some notes, then ^Z their vim session, then `kill %+` would kill that and leave an orphaned .swp file as well as leave them in the same position of having the port remain open.

dannysauer: A separate terminal would work fine too. That's common in tutorials, so people probably wouldn't think twice about it. There's something to be said for simplicity.

Should we want a bit more sophistication, since we _are_ the experts ;) - the `%kubectl` format I mentioned addresses the "wrong process" issue. It can contain the whole quoted command line if need be for further disambiguation, like `%"kubectl port-forward"` - which is also pretty descriptive of what's happening.

Either way, I still might argue that there could be a valuable lesson hidden here for the one person who reads our docs, backgrounds commands a lot, and blindly copy-pastes commands. 😂

dannysauer: Verbose rambling and apparent bias aside, I really do think both options are equivalent and both preferable to just leaving it.

cmurphy: Like I said, I used the `&` in keeping with the [upstream documentation](https://github.com/nvidia/gpu-monitoring-tools#quickstart-on-kubernetes) for this tool, not because I don't know how linux job control works. In looking for other examples online, I don't see anyone using `&`, they instead just opt to ignore the fact that the port-forward command takes over your terminal. I've changed it to remove the `&` and explicitly instruct the user to open a separate terminal. I think this is simplest and doesn't leave anyone guessing about the meaning of obscure kill syntax. I really hope we can leave it at this.

adoc/admin-gpus.adoc
nkoranova: This is a good an not trivial question actually. You are right about the loading time and we should really link to the `html` version rather than the `single-html`, though the loading time is not such a significant difference as our guides are still not that big. However, I prefer linking to the `single-html` version, because it makes it easier for me to work with the documents and probably about 95% of our really are to the `single-html` version. I'm going to create an issue to address this in our docs globally.

nkoranova: https://github.com/SUSE/doc-caasp/issues/927

adoc/admin-gpus.adoc
nkoranova: You can use `menu:File[Manage Dashboards]`. I'll go ahead and change that during the review. :)

PR #903, opened 2020-06-24
Update AWS example and fix stack naming

lcavajani:
# Describe your changes

This commits removes the hard-coded values in terraform.tfvars in favor of environment variables.

aws_az for availability zone is not implemented so it is removed as well.

It also sets a generic name for the stack without testing or specific reference to caasp version.

# Related Issues / Projects

Relates to: https://github.com/SUSE/skuba/pull/1190


nkoranova:
Thanks, looks like the Skuba PR was merged, so we can also merge this. @lcavajani is this also relevant to v4 as well as v5? Should we backport this to v4?

lcavajani:
@nkoranova I don't think this will be backported so just v5

Issue #820, opened 2020-05-26
[doc] 3.1.1  Basic SSH Key Configuration - key with ed25519 instead of ecdsa, timeout for ssh-agent


Martin-Weiss:
3.1.1  Basic SSH Key Configuration

https://susedoc.github.io/doc-caasp/master/single-html/caasp-deployment/#ssh.configuration

------------------------

Why do we recommend to use ed25519 instead of ecdsa? I thought ecdsa is the new sles default?

For loading the ssh-agent we should add a timeout to expire loaded keys for security reasons!

So could we change: 

`eval "$(ssh-agent)"`

to timeout 1 hour:

`eval "$(ssh-agent -t 3600)"`

r0ckarong:
@ereslibre @kkaempf Who can answer the question about the choice of cipher?

Do we have any technical reasons not to implement agent timeout?

kkaempf:
> @ereslibre @kkaempf Who can answer the question about the choice of cipher?
> 
> Do we have any technical reasons not to implement agent timeout?

@jsegitz should be best suited to answer security question within CaaSP.

jenting:
cc @dannysauer

dannysauer:
Sorry - I missed the tag. :)

ed25519 is slightly faster than ecdsa, but no one will actually notice that. They're both eleptic curve algorithms and quite good; mathematically they're essentially the same in terms of theoretical security, and both better than prime factor based algorithms in about every possible way.

The ecdsa signing algorithm (curve P-256) has been around longer (in ssh, especially) and is, as a result, more broadly compatible. That makes it a good general purpose default for the OS.  However, ed25519 has a number of subtle implementation advantages, including the implementation explicitly preventing various side channel attacks. http://ed25519.cr.yp.to/ describes it a bit more. There is also a curve comparison at https://safecurves.cr.yp.to/, if deeper reading or general reasons to argue with DJ Bernstein are desired. Everyone loves arguing with DJB. :D 

Curve 25519 is also not a US NIST-developed curve with tinfoil hat back doors like curve P-256, for people who suspect such unlikely things. ;)

Anyway, because the CAASP endpoints are going to both be the same version of SLES supporting the same algorithms, "broad comparability" is irrelevant. As such, I suggested ed25519 as the best signing algorithm to use when generating a key for CaaSP.

dannysauer:
[...]

PR #904, opened 2020-06-24
Add warning about ssh-agent

dannysauer:
# Describe your changes
Added more detail about ssh-agent risk and describe key timeouts in a warning box.  Not sure if that's a Warning or Tip or Note or something else, so I made a guess. :shrug:  I added it in both the VMWare and perparation sections, since both reference ssh-add.  More detail in the preparation doc, since that seemed likely to be most read.

I also tagged this with v4 and v5, though since it's a PR against master clearly I'm assuming v5.  I think we're doing one more 4.x release, so I'd be fine with backporting this to the v4 tree as well if so desired; it's not version-specific information, and should be easy enough to stab the extra section in.  I'll leave that call up to the people who know more about the docs. :)

# Related Issues / Projects

Relates to:
#820


nkoranova:
Thanks for the PR, I'm happy to backport this to v4. :)

PR #908, opened 2020-06-29
Update admission controller info in GPU guide

cmurphy:
In v5 skuba will enable the ExtendedResourceToleration admission
controller by default, so users will not have to manually enable it and
maintain it through upgrades. This patch removes these now-unnecessary
instructions, makes the discussion of the admission controller more
prominent, and adds further clarification on the effects of tainting
nodes.

# Related Issues / Projects

Relates to: https://github.com/SUSE/avant-garde/issues/1699
RFC: https://github.com/SUSE/caasp-rfc/pull/62
Skuba PR: https://github.com/SUSE/skuba/pull/1192

### DIFF
--- a/adoc/admin-gpus.adoc
+++ b/adoc/admin-gpus.adoc
@@ -86,41 +86,22 @@ In a heterogeneous cluster, it may be preferable to prevent scheduling pods that
 
 [source,bash]
 ----
-$ kubectl taint nodes worker0 nvidia.com/gpu=:NoSchedule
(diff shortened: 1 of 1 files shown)


cmurphy:
Using label "On-Hold" until the related RFC and skuba PR are merged.

jordimassaguerpla:
ExtendedResourceToleration admission controller is not enabled in 5.0.0 but will be in v5 lifetime. This should not be in the current 5.0.0 documentation.

PR #924, opened 2020-07-10
Add important note to etcd cluster restore

c3y1huang:
# Describe your changes
Some platform does not have public IP available from inside VMs and
will cause container crash loop during restore.

# Related Issues / Projects

Relates to: 
* https://github.com/SUSE/avant-garde/issues/1634
* https://github.com/SUSE/caasp-test-cases/pull/65

Signed-off-by: Chin-Ya Huang <chin-ya.huang@suse.com>


c3y1huang:
Moved to staging so QA and Docs can pick this up.

Issue #869, opened 2020-06-07
[doc] 4.2.3.2.2  vSphere In-tree CPI (VCP) - how to enable vsphere CPI on an existing cluster?


Martin-Weiss:
4.2.3.2.2  vSphere In-tree CPI (VCP):

https://susedoc.github.io/doc-caasp/master/single-html/caasp-deployment/#cluster.bootstrap.vcp

-----

I can not find how to enable the vsphere CPI on an existing cluster.. is that documentation available, somewhere?

r0ckarong:
@innobead

innobead:
We don't support for enabling vSphere CPI in the existing cluster due to some changes introduced to terraform script operation. 

Do we still have this request for now? 

cc @c3y1huang

Martin-Weiss:
> 
> 
> We don't support for enabling vSphere CPI in the existing cluster due to some changes introduced to terraform script operation.
> 
> Do we still have this request for now?
> 

Yes - all customers that are currently on VMware and that are using NFS because they do not have SES will have the need and motivation to move from NFS to VSPhere storage.

So at least we have to describe the manual steps required to enable VSphere Storage on existing CaaSP clusters in the documentation.
We also have to make sure that the VSphere CPI/CSI works well in case NO terraform is used to create the VMs.

innobead:
I have created an issue in the engineering backlog, and let's follow up there.

PR #930, opened 2020-07-17
Enable vSphere cloud provider on running cluster

c3y1huang:
# Describe your changes
Add procedures to enable vSphere cloud provider on a running cluster.

# Related Issues / Projects

Relates to:
* https://github.com/SUSE/doc-caasp/issues/869
* https://github.com/SUSE/avant-garde/issues/1814
* https://github.com/SUSE/caasp-test-cases/pull/68

Signed-off-by: Chin-Ya Huang <chin-ya.huang@suse.com>

### DIFF
--- a/adoc/deployment-bootstrap.adoc
+++ b/adoc/deployment-bootstrap.adoc
@@ -412,10 +412,12 @@ For installation instructions, refer to: https://github.com/vmware/govmomi/tree/
 You can use `govc` to automate the task.
 +
 [WARNING]
+====
 Setup `disk.enabledUUID` requires virtual machine to be powered off. The following script
 will setup all virtul machine in parallel, hense resulting some cluster downtimes while
 all machines are powered off. Modify the script or simply DO NOT use the script if minimal
 downtime is in consideration.
+====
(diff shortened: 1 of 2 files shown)


adoc/deployment-bootstrap.adoc
```diff
@@ -389,6 +390,144 @@ Clusters provisioned following link:{docurl}single-html/caasp-deployment/#_using
...
+  sudo systemctl start kubelet
+====
+
+. On every master node, update control-plane components.
++
+====
+sudo kubeadm upgrade node phase control-plane --etcd-upgrade=false
+====
+
+. On every worker node, update kubelet.
```
innobead, line 519: Is it duplicated as the previous step?

c3y1huang: Yes this is duplicated but one step in between. I've separated this intentionally so there is no mix procedures between masters and workers, and user doesn't have to switch back and forth.

innobead: Thanks for the explanation!

PR #952, opened 2020-07-30
Added a logging section

### LINKED ISSUE #827: [doc] 4.2.5  Cluster Bootstrap - add logging examples to skuba node join etc.
4.2.5  Cluster Bootstrap

https://susedoc.github.io/doc-caasp/master/single-html/caasp-deployment/#cluster.bootstrap

---------------------------
In the examples we should add `-v5 |tee <nodename>-skuba-join.log` or similar so that customers have a log of what ever happens in skuba and it is easier to help our customers in case they have issues..

mssola:
# Describe your changes

This pull request adds a common file that describes how we handle logging in skuba and how it can be used by our customers. I've written it as a common file because it might be useful in different guides, even though I've just included it in one of them (which is hopefully the proper one).

# Related Issues / Projects

Fixes #827


r0ckarong:
This is not really suitable as a "common_" include. This reads more like an introduction to our logging approach with skuba and can be re-written to intro the other logging topics. 

My suggestion would be to add the `| tee ...` pipe to the skuba examples in the bootstrap (closing the original bug) and then putting a shortened section about logging approaches in the deployment guide. Then move the contents of this PR into the logging section of the admin guide and pull together the logging topics there for a new chapter. Then we can link to this new chapter in the admin guide and have the logging concepts all in one place.

@mssola What do you think?

r0ckarong:
I completely reworked the logging section now and collected everything in one place. Backported to 4.2.2.

PR #956, opened 2020-08-03
Add information about addons and registries customization

r0ckarong:
Some information yet unclear. Which items can be updated this way? How would you update those using skuba across the cluster (if possible)?


r0ckarong:
@ereslibre Could you provide me with more information how this works or point me to the person who can?

ereslibre:
@r0ckarong regarding cri-o, @evrardjp or @mjura can provide information.

For addons, https://github.com/SUSE/caasp-rfc/blob/master/accepted/0046-addon-customizations.md was implemented, so the behavior should be the described there.

r0ckarong:
Replaced by #990

PR #960, opened 2020-08-06
Fix wording for module oddity

r0ckarong:
@jordimassaguerpla Could you please confirm that this is how it's going to work for the foreseeable future?

Thanks

Relates to https://github.com/SUSE/avant-garde/issues/1862


jordimassaguerpla:
I confirm.

PR #979, opened 2020-08-27
A guide for migration to CaaSP 4.5

atighineanu:
- This PR describes how can migration to CaaSP version 4.5 be carried out.
Relates to:
SUSE/avant-garde#1769
 SUSE/caasp-release-notes#51
 SUSE/avant-garde#1918


adoc/admin-migrate-caasp4.5.adoc
```diff
@@ -0,0 +1,74 @@
+== Migration to {productname} 4.5
+
+1) The OS must be connected to SCC. Otherwise the migration will not work.
```
kkaempf, line 3: Connecting hosts to the internet is the exception, not the rule in enterprise environments. How would this work with an RMT server ?

evrardjp: Well, I suppose in that case there is no automation, and people are setting up the repositories to their server. This is the same in here, so we might need to borrow SLE documentation, and just mention our repositories in addition ?

adoc/admin-migrate-caasp4.5.adoc
```diff
@@ -0,0 +1,74 @@
...
+2) You also need the new `zypper migration` plugin. This plugin is used to perform the needed tasks in order to migrate the node itself to the latest version of {platform}, such as updating the repositories to the new ones, and calling `zypper dup`. This plugin is provided by the `zypper-migration-plugin` package. Therefore, install the `zypper-migration-plugin` package:
+
+----
+zypper -n in zyppper-migration-plugin
+----
+
+3) Then, run the newly installed `zypper-migration` plugin (on the management node first, then on the rest of the nodes):
+
+----
+zypper migration --non-interactive --auto-agree-with-licenses
```
evrardjp, line 19: no need to do it non interactive and auto agree with licenses, in the docs.

adoc/admin-migrate-caasp4.5.adoc
```diff
@@ -0,0 +1,74 @@
...
+5) And now run the skuba cluster upgrade commands as it's done below.
+
+- First, check if there are any addons or components to upgrade before you upgrade the nodes:
+----
+skuba cluster upgrade plan
+skuba addon upgrade plan
+skuba addon upgrade apply
+----
+
+- Then, check with `cluster status` if all nodes have the same kubernetes version:
```
evrardjp, line 42: ... `which must be 1.17.x`

evrardjp (commented):
I would change a few things for clarity:

1) The zypper migration feature is for direct access to SCC as mentioned. It's not impossible to do without SCC, but I think it needs another section of documentation.
This should refer to the SLE docs IMO.

This is a prerequisite for the rest of the ugprade, which could be in yet a different section.

For example:
```
header1: Before the upgrade, migrate your repositories and upgrade to SP2
header1a: for SCC connected installs
header1b: with other installs
header2: Trigger the upgrades as usual
```

adoc/admin-migrate-caasp4.5.adoc
kkaempf: We should at least be explicit about migration with an RMT server.

kkaempf (changes requested):

r0ckarong:
> I would change a few things for clarity:
> 
> First, remove all the automatisation (licenses, non interactive).
> 
> Then, clarify that the zypper migration feature is for direct access to SCC as mentioned.
> It's not impossible to migrate and upgrade without SCC, but I think it needs more documentation, if possible in another section. This should refer to the SLE docs IMO.
> 
> As this SP migration is a prerequisite for the rest of the ugprade I think we could split the docs in two sections. I am not a docs architect, so I am not sure. I trust your experience y'all : )
> 
> Here is what I thought, for example:
> 
> ```
> header1: Before the upgrade, migrate your repositories and upgrade to SP2
> header1a: for SCC connected installs
> header1b: with other installs
> header2: Trigger the upgrades as usual
> ```

OK, so who can perform the testing for these scenarios and provide the resulting procedures?

adoc/admin-migrate-caasp4.5.adoc
r0ckarong: I thought 4.5.0 contains k8s 1.18?

adoc/admin-migrate-caasp4.5.adoc
kkaempf: Looking through [zypper-migration-plugin](https://github.com/SUSE/zypper-migration) it seems as if there's no difference between 'connected to SCC' and 'connected to RMT'. Reaching out to the author now for confirmation ...

adoc/admin-migrate-caasp4.5.adoc
atighineanu: > I thought 4.5.0 contains k8s 1.18?

That's before the upgrade takes place. The actual command that upgrades kubernetes to 1.18 is "skuba node upgrade apply --user sles --sudo --target IP"

atighineanu: > ... `which must be 1.17.x`

Will add that right now...

adoc/admin-migrate-caasp4.5.adoc
atighineanu: > no need to do it non interactive and auto agree with licenses, in the docs.

Actually that is true. Changing it.

adoc/admin-migrate-caasp4.5.adoc
atighineanu: I wasn't the person who tested airgapped clusters. I will have to ask around and/or do it myself again to make sure everything is in line with what is stated here.

kkaempf: Just got confirmation from on of the `zypper migration` authors that it works with SCC and RMT. (Event SMT would work but isn't supported with SLE15 afaik).

Bottom line: We can leave the `zypper migration` call and just need to remove the "SCC" parts.
It should be "registration server (like SCC or RMT)". This should cover 'airgap' scenarios well enough.

We should also point to the SLES upgrade guide at https://documentation.suse.com/sles/15-SP2/

evrardjp:
@r0ckarong this process already written _IS_ what has been tested.

adoc/admin-migrate-caasp4.5.adoc
```diff
@@ -1,11 +1,12 @@
 == Migration to {productname} 4.5
 
-1) The OS must be connected to SCC. Otherwise the migration will not work.
-In order to reconnect your OS to SCC, you run:
+1) The OS must be connected to servers like SCC or RMT. Otherwise the migration will not work.
```
kkaempf, line 3: The server must be ...

kkaempf: Scratch "Otherwise the migration will not work." and replace it with a reference to the SLES documentation. Like "Other migration scenarios are covered in the SLES upgrade guide"

adoc/admin-migrate-caasp4.5.adoc
```diff
@@ -1,11 +1,12 @@
...
 
-----
-SUSEConnect -r <your SCC key> SUSEConnect -p sle-module-containers/15.1/x86_64 -r <your SCC key>
-----
+[NOTE]
+====
+In order to reconnect your OS to SCC, you run:
```
kkaempf, line 7: "In order to reconnect your system to the registration server, run"

kkaempf (approved):
I'm fine with the content of the PR now. Some wording fixes remain.

evrardjp:
Agreing with @kkaempf comments.

thehejik:
Maybe we should add a note into this PR how to drain a node before performing manual `reboot` after `zypper migration` (which is not covered by `skuba-update` nor `kured`). But maybe simple `sudo touch /var/run/reboot-needed` would be sufficient for kured to reboot a node automatically after while.

kkaempf:
Good point ! Each node needs to be drained/cordoned before starting the migration. We had this discussion on RC before.

evrardjp:
> Good point ! Each node needs to be drained/cordoned before starting the migration. We had this discussion on RC before.

This is now included in skuba automatically.

thehejik:
> > Good point ! Each node needs to be drained/cordoned before starting the migration. We had this discussion on RC before.
> 
> This is now included in skuba automatically. But yes, it does not cover the case in which folks want to perform a manual reboot.

But we need to reboot right after `zypper migration` which is one step before calling skuba. Or do you mean that once I start `skuba node upgrade` it will first reboot the node and after that it will try to upgrade packages from v4.5?

Martin-Weiss:
Not sure if the full customers requirements are covered here, already.  

Please ensure that we can also do an upgrade with manually repo-changes (zypper ar / zypper rr) and using "zypper dup" instead of "zypper migration"! We have many enterprise customers that do NOT register their servers to SCC/SMT/RMT and have their own repo-sources (ZCM, SUSE Manager). In these cases we have to adjust the repos manually and us zypper dup...

kkaempf:
> Please ensure that we can also do an upgrade with manually repo-changes (zypper ar / zypper rr) and using "zypper dup" instead of "zypper migration"!

Is this a supported scenario for SLES migrations ?

In any case, there's not too much we can document for zypper repos, except
- switch your SLE repos from 15 SP1 to 15 SP2 (incl. containers module for nodes and public cloud for the management/skuba host)
- switch from caasp-4.2 to caasp-4.5

Martin-Weiss:
> 
> 
> > Please ensure that we can also do an upgrade with manually repo-changes (zypper ar / zypper rr) and using "zypper dup" instead of "zypper migration"!
> 
> Is this a supported scenario for SLES migrations ?
> 

AFAIK - yes - we have to do this in SUMA assisted deployments often and we have to do this with SES, too..

> In any case, there's not too much we can document for zypper repos, except
> 
>     * switch your SLE repos from 15 SP1 to 15 SP2 (incl. containers module for nodes and public cloud for the management/skuba host)
> 
>     * switch from caasp-4.2 to caasp-4.5

Documentation is fine - I believe we just need to list the required repos properly..

r0ckarong:
@atighineanu I do not really understand the purpose of the repos list. It's just the names and aliases. It says nothing about any status change between migration states. It also doesn't allow to verify the URI (which contains the version).

Can we just have a list with the repo URIs that need to be enabled after migration? If not, what benefit does the user have from seeing more than a dozen _disabled_ repositories?

r0ckarong:
@kkaempf @Martin-Weiss From my side the requested information for this step is there now. Please give me a thumbs up and I will merge.

kkaempf (approved):

evrardjp:
@r0ckarong  I just realised that the upgrade of the local config for cri-o is only in release notes. Shouldn't it be also here, in the process? (skuba cluster upgrade localconfig is documented in https://github.com/SUSE/caasp-release-notes/pull/51/files )

Martin-Weiss:
I thought "addon upgrade apply" just deploys yaml and does not need ssh? Do we control the crio config via yaml and can skuba distribute the crio.conf.d changes with addon apply and without ssh?

evrardjp:
correct it's not automatically done.
The reason is that it doesn't happen at the same time as the addon upgrade.

For example, for version 1.17.9, the addons are at version x, and for 1.18.4, the addons are at version y (fictional numbers).
When following the process, the skuba addon upgrade apply is first done, which applies addon upgrades from "x" (1.17.9). It's a post upgrade case. You then upgrade your cluster, and generally we had no action to do (as things didn't need manual intervention, we could auto migrate things). However, for 1.18.4, this is not the case, as y is not backwards compatible with x. Y migration needs to happen _just before_ upgrading your node _BUT_ needs manual intervention/checking from the deployer (which prevents us from doing things automatically). This is why this command appear. You should check with the contributors of this command, and the card implementing it. @mssola can help there.

evrardjp:
> I thought "addon upgrade apply" just deploys yaml and does not need ssh? Do we control the crio config via yaml and can skuba distribute the crio.conf.d changes with addon apply and without ssh?

1) It's too late to change the design ;)
2) crio.config is _not_ managed through yaml, but is manually managed in the cluster definition folder of the users, like all other addons. A user can change it manually and it will automatically get applied, like it used to be done. However, this is by far not convenient to say to someone "Please rename those files, add another one there, and you are ready", while we can simply provide a command that is smart and does the job.

Martin-Weiss:
1. then we need to fix the bug in the design ASAP ;-)
2. we need a solution for customers and make it easy for them to upgrade from 4.x to 4.5 that also takes care about their existing customizations of crio.conf (especially the global_auth_file). At the moment we do NOT have a working and "easy" solution as the solution is not implemented "fully automated" nor "documented" in a way a customer can follow.

I would recommend to document the step by step process properly and hopefully get this fixed in one of the next releases.
(this is the third of fourth time where we have to struggle with crio.conf since CaaSP 4.0)

r0ckarong:
So we are missing the step of refreshing the config manually during the migration. What is the correct way to do this and at what point in the current instructions does this need to be done?

@atighineanu Did you guys test with a modified crio.conf?

evrardjp:
> we need a solution for customers and make it easy for them to upgrade from 4.x to 4.5 that also takes care about their existing customizations of crio.conf (especially the global_auth_file). At the moment we do NOT have a working and "easy" solution as the solution is not implemented "fully automated" nor "documented" in a way a customer can follow.

It is, that's indeed the scope of that command I mentioned above.

evrardjp:
@r0ckarong I think @atighineanu indeed tested with a modified crio, else it wouldn't work ;)
We just forgot to document it. But I will let @atighineanu speak, instead of talking for him :p

evrardjp:
What was tested _when developing_ was only the features that were enabled through feature flags. Manually overriden data still need manual intervention, this is why we don't automate everything. We can't assume what people have done with their config, as it would be quite a gammut of changes. We should just encourage deployers to try the automation tool to migrate to the new configuration structure, and kindly ask them to review the changes.

evrardjp:
> 1. then we need to fix the bug in the design ASAP ;-)

I agree on that: if there is something we should iterate, let's iterate! Let's just not redesign from scratch.

PR #1019, opened 2020-09-30
Remove manual certificate renewal

jenting:
# Describe your changes

we already have auto certificate renewal solution `cert-manager`, remove
the manual way from the admin guide.

# Related Issues / Projects

Relates to:


innobead (approved):
:+1:

r0ckarong:
@innobead @jenting What is the recommended method to fix a cert if the automatic renewal won't work? Shouldn't we move the manual method to troubleshooting instead of completely removing it?

jenting:
Hmmm... you're right, then we missing part of certificates on how to rotates it, close this PR now and adding another PR for the missing part.

PR #1024, opened 2020-10-05
Add instruction on how to auto rotate cilium etcd client cert (bsc#1176904)

jenting:
# Describe your changes

Add instruction on how to auto rotate cilium etcd client cert.

# Related Issues / Projects

Relates to: https://github.com/SUSE/avant-garde/issues/1969

### DIFF
--- a/adoc/admin-security-certificates.adoc
+++ b/adoc/admin-security-certificates.adoc
@@ -678,8 +678,148 @@ The kubelet client certificate are signed by kube-controller-manager and the kub
 
 The addon certificates can be automatically rotated by leveraging the functions of the open-source solutions `cert-manager` and `reloader`. `cert-manager` is for automatically rotating certificates stored in Secrets, and `reloader` is for watching and reconciling the updated Secrets to execute a rolling upgrade of the affected Deployments or DaemonSet.
 
-* Prerequisites
+. Install reloader via helm chart:
++
+[source,bash]
+----
+helm install \
+    --name <RELEASE_NAME> \
+    --namespace cert-manager \
+     suse/reloader
+----
++
+Or if you have selected the Helm 3 alternative also see <<helm-tiller-install>>:
++
+[source,bash]
+----
+helm install <RELEASE_NAME> \
+    --namespace cert-manager \
+    --create-namespace \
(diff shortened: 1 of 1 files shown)


adoc/admin-security-certificates.adoc
```diff
@@ -678,72 +678,183 @@ The kubelet client certificate are signed by kube-controller-manager and the kub
...
+The cert-manager has a custom resource, Certificate, which can be used to define a requested x509 certificate which will be renewed and kept up to date by an Issuer or ClusterIssuer resource.
+
+==== Client Certificate Rotation
+
+[WARNING]
+====
+If you are running a cluster using cilium version before 1.6, the cilium data is stored in the ETCD cluster, not the custom resources (CR). {skuba} generates a client certificate to read/write the cilium date to the ETCD cluster and the client certificate will expire after 1 year. Please follow the below steps to use cert-manager to automatically renew the cilium client certificate.
+====
 
+. Check the {productname} cilium version before 1.6
```
innobead, line 747: Do you mean before cilium addon upgrade, people need to check if their cilium is before 1.6, then set up the cert rotation beforehand?

Do we need to consider some statements for after cilium addon upgrade (1.5 -> 1.6) but cilium not working because the migrated cilium-secret does have expired certs? So that means not just check if cilium before 1.6, also need to check if the version is under 1.6 but the cluster not working?

innobead (commented):
All good, just need to clarify the statements about cilium secret before or after cilium addon upgrade.

adoc/admin-security-certificates.adoc
jenting: Before cilium < 1.6, cilium stores data into the ETCD cluster, cilium >= 1.6, cilium stores data into custom definition (CR).
So the admin _only_ needs to rotate the cilium client certificate if cilium < 1.6.

Do I need to add other sentences to make it more clear?

innobead: It was clarified now! Thanks :+1:

maximenoel8:
After testing with cilium 1.15. The steps of the documentation solved the certificate issue.

